PR #57, opened 2017-07-21
FIX the caching

teonbrooks:
The caching is leading to numerical inaccuracies. This reverts the function


teonbrooks:
closing because we're working directly on master

PR #90, opened 2018-03-26
[WIP][API]: clean up and happily expose .get_reject_log

dengemann:
Next attempt @jasmainak @agramfort

### DIFF
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -393,7 +393,8 @@ def compute_thresholds(epochs, method='bayesian_optimization',
     return threshes
 
 
-def _apply_log(local_reject, epochs, annot, threshes_, picks_, verbose):
+def _log_apply_interp(local_reject, epochs, annot, threshes_, picks_,
+                      verbose):
     if not all(epochs.ch_names[pp] in threshes_
                for pp in picks_):
         raise ValueError('You are passing channels which were not present '
@@ -404,6 +405,15 @@ def _apply_log(local_reject, epochs, annot, threshes_, picks_, verbose):
         local_reject._interpolate_bad_epochs(
             epochs, interp_channels=annot['interp_channels'][ch_type],
             picks=this_picks, verbose=verbose)
+
+
+def _log_apply_drop(local_reject, epochs, annot, threshes_, picks_,
+                    verbose):
+    if not all(epochs.ch_names[pp] in threshes_
+               for pp in picks_):
+        raise ValueError('You are passing channels which were not present '
+                         'at fit-time. Please fit it again, this time '
+                         'correctly.')
 
     if np.any(annot['bad_epochs_idx']):
         epochs.drop(annot['bad_epochs_idx'], reason='AUTOREJECT')
@@ -554,13 +564,13 @@ class LocalAutoReject(BaseAutoReject):
 
         return bad_epochs_idx, sorted_epoch_idx, n_epochs_drop
 
-    def _get_reject_log(self, threshes, epochs, picks):
+    def _get_reject_log(self, epochs, picks):
         """Get essential annotations for epochs given thresholds."""
         ch_type = _get_ch_type_from_picks(
             picks=picks, info=epochs.info)[0]
 
         drop_log, bad_sensor_counts = self._vote_bad_epochs(
-            epochs, threshes=threshes, picks=picks)
+            epochs, picks=picks)
         interp_channels, fix_log = self._get_epochs_interpolation(
             epochs, drop_log=drop_log, ch_type=ch_type, picks=picks,
             n_interpolate=self.n_interpolate_[ch_type])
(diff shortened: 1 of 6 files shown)


dengemann:
Not to be reviewed. I first got to talk to @jasmainak to converge on return API for annotate_epochs.

autoreject/viz.py
```diff
@@ -160,7 +160,7 @@ def plot_epochs(epochs, picks=None, scalings=None, n_epochs=20,
         Defaults to False.
     bad_epochs_idx : array-like | None
         Indices of bad epochs to show. No bad epochs to visualize if None.
-    fix_log : array, shape (n_channels, n_epochs) | None
+    log_labels : array, shape (n_channels, n_epochs) | None
```
jasmainak: `bad_labels`?

jasmainak:
How far are we from seeing this PR merged? Do you think we could break this PR down into small mergeable chunks?

autoreject/autoreject.py
```diff
@@ -469,34 +454,32 @@ def _vote_bad_epochs(self, epochs):
...
-        non_picks = np.setdiff1d(range(epochs.info['nchan']), self.picks)
-        interp_channels = list()
-        n_interpolate = self.n_interpolate[ch_type]
+        labels = drop_log.copy()
+        non_picks = np.setdiff1d(range(epochs.info['nchan']), picks)
         for epoch_idx in range(len(epochs)):
-            n_bads = drop_log[epoch_idx, self.picks].sum()
+            n_bads = drop_log[epoch_idx, picks].sum()
             if n_bads == 0:
-                interp_channels.append([])
```
jasmainak: this needs to be fixed

autoreject/autoreject.py
```diff
@@ -614,69 +621,120 @@ def transform(self, epochs):
         ----------
         epochs : instance of mne.Epochs
             The epochs object for which bad epochs must be found.
+
+        return_log : bool
+            If true the rejection log is also returned.
```
jasmainak: If True

autoreject/autoreject.py
```diff
@@ -614,69 +621,120 @@ def transform(self, epochs):
...
+
+        Returns
+        -------
+        epochs_clean : instance of mne.Epochs
+            The cleaned epochs.
+
+        reject_log : instance of autoreject.RejectLog
+            The rejection log. Returned only of return_log is True.
```
jasmainak: only if

examples/plot_visualize_bad_epochs.py
```diff
@@ -146,16 +153,13 @@
...
-            title='')
+scalings = dict(eeg=40e-6)
+reject_log.plot_epochs(this_epoch, scalings=scalings)
 
 ###############################################################################
 # ... and the epochs after cleaning with autoreject
 
-epochs_ar.plot(scalings=dict(eeg=40e-6))
-
+epochs_ar.plot(scalings=scalings)
```
jasmainak: personally I have struggled with this scalings parameter in MNE. It would be nice to have scalings='auto' or something like that without having to set it individually for each channel type.

dengemann:
Pretty unrealistic. There is a bug that needs to be fixed. Then we can
address cosmetics.

dengemann:
Yeah. Keep in mind the goal: settle api for pip release. We can develop
more later.

dengemann:
Yeah why not. Lables was already taken in viz code. So I added some prefix
ad-hoc.

autoreject/autoreject.py
dengemann: What exactly? It is obsolete now. We extract the bad interp chans from fix log.

jasmainak: oh I see. Makes sense

jasmainak:
@dengemann let's be brave and try a rebase here? ;)

jasmainak:
@dengemann? What is the status here? Do you need help? I can take over if needed

agramfort:
please take over @jasmainak

dengemann:
I have still no time to get back to this. Perhaps later tonight. Please
take over if you have time!

dengemann:
/ push into branch.

autoreject/tests/test_autoreject.py
```diff
@@ -127,62 +131,68 @@ def test_autoreject():
...
 
     # test complementarity of goods and bads
-    assert_array_equal(
-        np.sort(np.r_[bad_epochs_idx, good_epochs_idx]),
-        np.arange(len(epochs_fit)))
+    assert_array_equal(len(reject_log.bad_epochs), len(epochs_fit))
 
     # test that transform does not change state of ar
-    epochs_fit.fit_ = True
     epochs_clean = ar.transform(epochs_fit)  # apply same data
```
jasmainak: @dengemann this test is failing for some reason. `ar.picks_` is broken ... it's also the case in the non-rebased PR

PR #93, opened 2018-04-09
WIP: Fix transform with bad channels.

mattboggess:
This is the fix for the issue raised about transform interpolating bad channels back into the data.

I've added a test by comparing autoreject run on data with and without extra bad channels. However, I'm running into a bug that I don't quite understand while trying to pass this test. Autoreject on the copy of the data with bad channels fails on an assertion in line 273 of _interpolate_bads_meg_fast. This assertion seems to be asserting that the bad channels in the sub-selected info and the original info should be the same? I'm not sure why this would be the case. Am I misunderstanding or is this another bug?

### DIFF
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -664,7 +664,7 @@ def _interpolate_bad_epochs(
             desc='Repairing epochs',
             position=pos, leave=True, verbose=verbose):
         epoch = epochs[epoch_idx]
-        epoch.info['bads'] = interp_chs
+        epoch.info['bads'] += interp_chs
         interpolate_bads(epoch, picks=picks, reset_bads=True)
         epochs._data[epoch_idx] = epoch._data
 
@@ -742,36 +742,6 @@ def _run_local_reject_cv(epochs, thresh_func, picks_, n_interpolate, cv,
                 loss[idx, jdx, fold] = -local_reject.score(X[test])
 
     return local_reject, loss
-            (_, _, interp_channels, _,
-             bad_epochs_idx, good_epochs_idx) = self._annotate_epochs(
-                 threshes=self.threshes_, epochs=epochs)
-            if len(good_epochs_idx) == 0:
-                raise ValueError('All epochs are bad. Sorry.')
-
-            self._interpolate_bad_epochs(
-                epochs_out, interp_channels=interp_channels,
-                verbose=self.verbose)
-        if np.any(bad_epochs_idx):
-            epochs_out.drop(bad_epochs_idx, reason='AUTOREJECT')
-        else:
-            warnings.warn(
-                "No bad epochs were found for your data. Returning "
-                "a copy of the data you wanted to clean. Interpolation "
-                "may have been done.")
-        return epochs_out
-
-    def _interpolate_bad_epochs(
-            self, epochs, interp_channels, verbose='progressbar'):
-        """Actually do the interpolation."""
-        pos = 4 if hasattr(self, '_leave') else 2
-        for epoch_idx, interp_chs in _pbar(
-                list(enumerate(interp_channels)),
-                desc='Repairing epochs',
-                position=pos, leave=True, verbose=verbose):
-            epoch = epochs[epoch_idx]
-            epoch.info['bads'] += interp_chs
-            interpolate_bads(epoch, picks=self.picks, reset_bads=True)
-            epochs._data[epoch_idx] = epoch._data
 
 
 class LocalAutoRejectCV(object):
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -664,7 +664,7 @@ def _interpolate_bad_epochs(
             desc='Repairing epochs',
             position=pos, leave=True, verbose=verbose):
         epoch = epochs[epoch_idx]
-        epoch.info['bads'] += interp_chs
+        epoch.info['bads'] = interp_chs
         interpolate_bads(epoch, picks=picks, reset_bads=True)
         epochs._data[epoch_idx] = epoch._data
 
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -742,6 +742,36 @@ def _run_local_reject_cv(epochs, thresh_func, picks_, n_interpolate, cv,
                 loss[idx, jdx, fold] = -local_reject.score(X[test])
 
     return local_reject, loss
+            (_, _, interp_channels, _,
+             bad_epochs_idx, good_epochs_idx) = self._annotate_epochs(
+                 threshes=self.threshes_, epochs=epochs)
+            if len(good_epochs_idx) == 0:
+                raise ValueError('All epochs are bad. Sorry.')
+
+            self._interpolate_bad_epochs(
+                epochs_out, interp_channels=interp_channels,
+                verbose=self.verbose)
+        if np.any(bad_epochs_idx):
+            epochs_out.drop(bad_epochs_idx, reason='AUTOREJECT')
+        else:
+            warnings.warn(
+                "No bad epochs were found for your data. Returning "
+                "a copy of the data you wanted to clean. Interpolation "
+                "may have been done.")
+        return epochs_out
+
+    def _interpolate_bad_epochs(
+            self, epochs, interp_channels, verbose='progressbar'):
+        """Actually do the interpolation."""
+        pos = 4 if hasattr(self, '_leave') else 2
+        for epoch_idx, interp_chs in _pbar(
+                list(enumerate(interp_channels)),
+                desc='Repairing epochs',
+                position=pos, leave=True, verbose=verbose):
+            epoch = epochs[epoch_idx]
+            epoch.info['bads'] += interp_chs
+            interpolate_bads(epoch, picks=self.picks, reset_bads=True)
+            epochs._data[epoch_idx] = epoch._data
 
 
 class LocalAutoRejectCV(object):
--- a/autoreject/tests/test_autoreject.py
+++ b/autoreject/tests/test_autoreject.py
@@ -90,13 +90,13 @@ def test_autoreject():
     pick_ch_names = [epochs.ch_names[pp] for pp in pre_picks]
     bad_ch_names = [epochs.ch_names[ix] for ix in range(len(epochs.ch_names))
 		    if ix not in pre_picks]
-    epochs_wbads = epochs.copy()
-    epochs_wbads.info['bads'] = bad_ch_names
+    epochs_with_bads = epochs.copy()
+    epochs_with_bads.info['bads'] = bad_ch_names
     epochs.pick_channels(pick_ch_names)
 
     epochs_fit = epochs[:12]  # make sure to use different size of epochs
     epochs_new = epochs[12:]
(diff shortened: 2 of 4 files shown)


jasmainak:
Honestly I am a bit fuzzy on the picking part of the code.

Maybe @dengemann can help since he wrote it

autoreject/tests/test_autoreject.py
```diff
@@ -88,8 +88,13 @@ def test_autoreject():
         mne.pick_types(epochs.info, meg=False, eeg=True)[::16],
         mne.pick_types(epochs.info, meg=False, eeg=False, eog=True)]
     pick_ch_names = [epochs.ch_names[pp] for pp in pre_picks]
+    bad_ch_names = [epochs.ch_names[ix] for ix in range(len(epochs.ch_names))
+		    if ix not in pre_picks]
+    epochs_wbads = epochs.copy()
```
jasmainak: you mean `epochs_with_bads`? It could also be `epochs_without_bads`. Better to be explicit I guess

mattboggess:
Hey just wanted to see if there was any update on this.

dengemann:
Hi Matt. I‘ll look into it asap latest by Wednesday. You‘re officially
authorized to reping me till it hurts ...

jasmainak:
@mattboggess can you try rebasing your PR to see if this is still an issue in `master`?

mattboggess:
Sorry I messed up merge on first commit, but it's fixed now. Also added a few updates based on code changes and made your style edit suggestions.

It still seems to be failing in the same way though.

mattboggess:
@dengemann Were you able to look into this? Rebasing still ran into the same issue. Not sure how to proceed to close this.

autoreject/autoreject.py
```diff
@@ -664,7 +664,7 @@ def _interpolate_bad_epochs(
             desc='Repairing epochs',
             position=pos, leave=True, verbose=verbose):
         epoch = epochs[epoch_idx]
-        epoch.info['bads'] = interp_chs
+        epoch.info['bads'] += interp_chs
```
jasmainak: okay thinking more about this. I think the right fix is to use `picks` to ignore the user marked bad channel. The user is free to interpolate later if they want. This will affect also the estimation phase in `autoreject`.

jasmainak:
@mattboggess sorry for the late response. I have been traveling with no laptop.

Can you try to make it work with `picks`? I am sorry for the shifting requirement. This is what we intended to do. Basically `ar.fit()` and `ar.transform()` should both not touch the bad channel marked by user unless they explicitly ask for it using the `picks` argument. Does that make sense?

jasmainak:
>  Autoreject on the copy of the data with bad channels fails on an assertion in line 273 of _interpolate_bads_meg_fast. This assertion seems to be asserting that the bad channels in the sub-selected info and the original info should be the same? I'm not sure why this would be the case. Am I misunderstanding or is this another bug?

Feel free to edit it as long as you can make travis happy ...

mattboggess:
Thanks! That was definitely the right way to do it. It turns out this was an EEG specific bug where the interpolation code did not take into account the picks supplied.

agramfort:
@jasmainak ok for you?

autoreject/tests/test_autoreject.py
```diff
@@ -195,6 +201,24 @@ def test_autoreject():
     if not np.isscalar(is_same):
         is_same = np.isscalar(is_same)
     assert_true(not is_same)
+
+    # test that transform ignores bad channels
+    picks_without_bads = mne.pick_types(
+        epochs_with_bads.info, meg='mag', eeg=True, stim=False, eog=False,
+        include=[], exclude='bads')
```
jasmainak: this shouldn't be needed @mattboggess. See this: https://github.com/autoreject/autoreject/blob/master/autoreject/utils.py#L67-L68. `_handle_picks` should have taken care of the bad channel because it is called at the beginning of every function, no? If you use `picks=None`, if it does not ignore the bad channel then there is a bug. But we should make it work as before :-)

autoreject/tests/test_autoreject.py
```diff
@@ -195,6 +201,24 @@ def test_autoreject():
...
+                           thresh_func=thresh_func,
+                           n_interpolate=[1, 2],
+                           consensus=[0.5, 1])
+    ar.fit(epochs_with_bads_fit)
+    epochs_with_bads_clean = ar.transform(epochs_with_bads_fit)
+
+    epochs_with_bads_clean.pick_types(meg=True, eeg=True, eog=True,
+                                      exclude='bads')
+    assert_array_equal(epochs_clean.get_data(),
+                       epochs_with_bads_clean.get_data())
```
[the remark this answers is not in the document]
dengemann: Agreed.

autoreject/tests/test_autoreject.py
mattboggess: The bads should be ignored with picks=None as well. Nothing I edited should have changed this since _handle_picks will just replace the picks before all of my changes. I use the picks here to match the clean data generated from the previous tests for comparison with and without bad channels. Do you want me to add a separate test for picks=None?

autoreject/utils.py
```diff
@@ -214,8 +214,17 @@ def interpolate_bads(inst, picks, reset_bads=True, mode='accurate'):
     # emerges
     verbose = mne.set_log_level('CRITICAL', return_old_level=True)
 
-    # this needs picks, assume our instance is complete and intact
-    _interpolate_bads_eeg(inst)
+    eeg_picks = set(pick_types(inst.info, meg=False, eeg=True, exclude=[]))
```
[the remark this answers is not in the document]
mattboggess: The next line sub-selects only those eeg channels that were in the original picks supplied  This is what actually gets used for the interpolation (the eeg_interp_picks variable). See the newest edit for how this mirrors the MEG interpolation code.

autoreject/utils.py
```diff
@@ -214,8 +214,17 @@ def interpolate_bads(inst, picks, reset_bads=True, mode='accurate'):
...
+    eeg_picks_interp = [p for p in picks if p in eeg_picks]
+    if len(eeg_picks_interp) > 0:
+        keep_chs = [inst.ch_names[i] for i in eeg_picks_interp]
+        inst_interp = inst.copy().pick_channels(keep_chs)
```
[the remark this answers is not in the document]
mattboggess: That's a good point. I copied over MNE's EEG interpolation function and exposed a picks parameter. Is this an acceptable way to do it?

autoreject/tests/test_autoreject.py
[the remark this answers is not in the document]
mattboggess: Good idea. I added a second test that checks the bad channel data remains unchanged.

mattboggess:
@jasmainak Let me know if this new commit addresses your review sufficiently. As for the bad channels warning, where did this used to be? Looking at the autoreject master branch, the warning you reference is in the _check_data function. This is still there in my version and gets printed when I run the tests. Did you want me to add a similar warning to the interpolation function? I don't see it in the interpolation function on the autoreject master branch.

autoreject/tests/test_autoreject.py
```diff
@@ -195,6 +201,30 @@ def test_autoreject():
     if not np.isscalar(is_same):
         is_same = np.isscalar(is_same)
     assert_true(not is_same)
+
+    # test that transform ignores bad channels
+    picks_without_bads = mne.pick_types(
+        epochs_with_bads.info, meg='mag', eeg=True, stim=False, eog=False,
+        include=[], exclude='bads')
+    ar_bads = LocalAutoRejectCV(cv=3, picks=picks_without_bads,
```
jasmainak: this test is "trivial" if the picks is working, no? That is to say, if I pass any arbitrary integer list (with or without bad channels), it should just respect the `picks` that the user provides. I would be surprised if this failed on master. The case of `picks=None` is different as it is modified inside `autoreject` ...

autoreject/utils.py
```diff
@@ -229,6 +231,61 @@ def interpolate_bads(inst, picks, reset_bads=True, mode='accurate'):
     return inst
 
 
+def _interpolate_bads_eeg(inst, picks, verbose=None):
```
jasmainak: I would add a test to check if it gives the same result as MNE ... at least when `picks` is None

jasmainak:
> Looking at the autoreject master branch, the warning you reference is in the _check_data function. 

ah yes, you are correct. My bad!

autoreject/utils.py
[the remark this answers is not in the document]
jasmainak: yeah that's better!

autoreject/tests/test_autoreject.py
mattboggess: It fails on master because the EEG interpolation code didn't respect the picks. The previous code reset the user specified bads to good channels to be used for interpolation instead of excluding them from the interpolation process. This test directly tests for this bug by checking that the presence of bad EEG channels in the data doesn't change the outcome of autoreject on the good channels. So I don't think it's necessarily trivial given that the picks the user provided weren't being fully respected before. I agree that it'd be better to check the handling of picks=None in addition so I updated the test to check this. Does this seem reasonable?

autoreject/tests/test_autoreject.py
```diff
@@ -88,9 +88,15 @@ def test_autoreject():
         mne.pick_types(epochs.info, meg=False, eeg=True)[::16],
         mne.pick_types(epochs.info, meg=False, eeg=False, eog=True)]
     pick_ch_names = [epochs.ch_names[pp] for pp in pre_picks]
+    bad_ch_names = [epochs.ch_names[ix] for ix in range(len(epochs.ch_names))
+                    if ix not in pre_picks]
```
jasmainak: This looks weird Shouldn't the bad channels be a subset of the channels in the epochs? Right now, `bad_ch_names` is not a subset of `epochs.info['ch_names']`

jasmainak:
@mattboggess can you fix my last comment? Otherwise LGTM

autoreject/tests/test_autoreject.py
mattboggess: bad_ch_names is a subset of epochs_with_bads.info['ch_names']. epochs_with_bads and epochs have the same good channels, but I mark the channels that are dropped in epochs as bad in epochs_with_bads rather than drop them. This way I can compare autoreject on the same set of good channels with and without other bad channels in the epochs to make sure the bad channels do not affect the result.

jasmainak: yes, understood. I think I was just surprised by 360 bad channels :-)

jasmainak:
ready to merge @mattboggess @agramfort @dengemann ?

mattboggess:
Good on my end!

jasmainak:
Thanks @mattboggess !

agramfort:
Are bad channels ignored low but forced as globally bad and eventually interpolated by autorejct ?

jasmainak:
Bad channels marked by users should be ignored completely by autoreject. If they want, they can just do one more line and interpolate the bad channels:

```py
>>> epochs.interpolate_bads()
```

agramfort:
ok. It should be clearly documented.

jasmainak:
added in https://github.com/autoreject/autoreject/pull/101

PR #138, opened 2019-04-25
Fix n_jobs > 1 error

LegrandNico:
First attempt to fix the random sampling error when using Ransac with n_jobs > 1

### DIFF
--- a/autoreject/ransac.py
+++ b/autoreject/ransac.py
@@ -19,18 +19,12 @@ from .utils import _pbar, _handle_picks
 from .utils import _check_data
 
 
-def _iterate_epochs(ransac, epochs, idxs, verbose):
-
-    # Select random subset
-    if ransac.n_jobs > 1:
-        ransac.random_state = ransac.random_state + idxs.max()
-    ransac.ch_subsets_ = ransac._get_random_subsets(epochs.info)
-    ransac.mappings_ = ransac._get_mappings(epochs)
-
+def _iterate_epochs(ransac, epochs, idxs, ch_subset, verbose):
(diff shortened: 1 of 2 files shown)


LegrandNico:
see #136

autoreject/ransac.py
```diff
@@ -20,6 +20,13 @@
 
 
 def _iterate_epochs(ransac, epochs, idxs, verbose):
+
+    # Select random subset
+    if ransac.n_jobs > 1:
```
agramfort: it would be great to have the same code  path and results for any n_jobs.
Can you see how to obtain this? Precompute the seeds or random sample to pass to subprocesses?

jasmainak: @LegrandNico did you have time to address this comment? otherwise I'll take over

LegrandNico: Sorry, I did not have the time + I am not at ease with Joblib. I think this would require to change several parts of the script. Still, I am interested to see how you would do that.

jasmainak:
@agramfort I addressed the comment. Is it good now?

@LegrandNico can you update the whats_new.rst page to reflect your contribution? Thanks

jasmainak:
Merged by rebase after cosmit. Thanks @LegrandNico for the fixes. 

So happy about the CircleCI improvements @sappelhoff . It will make my job easier to make a release :-)

sappelhoff:
Makes me happy as well :)

PS: this merge should have closed https://github.com/autoreject/autoreject/issues/136 right?

jasmainak:
Indeed, thanks for pointing that out!

PR #147, opened 2019-06-05
[MRG] Extend documentation

### LINKED ISSUE #144: DOC: Provide "intuition" on what autoreject does
The paper on autoreject explains most of the details to understand the code. However, it's not a very easy read and I think that the docs would benefit from a small section providing an "intuition" on how autoreject works.

A possible place could be in the [FAQ](http://autoreject.github.io/faq.html).

Below I paste a first draft of what I envision such a "intuition" to be like for autoreject global.

Feel free to add comments on that text in this [Gdoc](https://docs.google.com/document/d/1jU6NleGgNvW2qcCh41Qt0WTBnaDntqDjY_5w6WEikm8/edit?usp=sharing)

---

autoreject global
=================
- Given some data with multiple trials (epochs) that contain sensor X timepoints data
- We want to find a threshold T in microvolts that will reject noisy epochs and retain clean epochs
- Do the following for a set of possible candidate thresholds Ts
- For each T:
    - Split your data into K folds (=k equal parts) along the trial dimension
        - Each of the k parts will be a "test" set once, while the others will be combined to be a "train" set
    - Then for each fold k (consisting of train and test trials) do:
        - apply threshold T to reject trials in the train set
        - calculate the mean of the signal (for each sensor and timepoint) over the GOOD trials in the train set
        - calculate the *median* of the signal (for each sensor and timepoint) over ALL trials in the test set
        - compare both of these signals and calculate the error (frobenius norm of their difference)
        - save the error
    - Now we have k errors
    - Form the mean error (over all k errors) associated with our current threshold T in microvolts
    - Save the mapping of current T to its error

- ... now we have a set of Ts all mapped to their specific error 
- the T with the lowest error is the best rejection threshold for a global rejection

sappelhoff:
closes #144 

Based on our discussion in #144 I try to put the "intuition" on autoreject global into a new section of the docs.

I think that this `explanations` section can grow over time and that it is a bit like the last "missing" part, when considering this [seminal pycon australia talk](https://www.divio.com/blog/documentation/) (link leads to associated blogpost)

image

I see:
- "learning oriented" covered by /examples
- "Problem oriented" covered by FAQ
- "understanding oriented" covered through this PR
- "information oriented" covered by the API

Note: The actual text in `explanation.rst` is still WIP, but I wanted your input on the structure before continuing to work on it.

### DIFF
--- a/doc/explanation.rst
+++ b/doc/explanation.rst
@@ -1,0 +1,35 @@
+Explanations
+============
+
+This section of the documentation exists to illuminate how ``autoreject`` works.
+The primary source for understanding should be the original publication [1]_,
+however the sections in this guide can make the content of that primary source
+more graspable.
+
+
+Intuition on how the - *autoreject global* - algorithm works
+------------------------------------------------------------
+- Given some data with multiple trials (epochs) that contain sensor X timepoints data
+- We want to find a threshold T in microvolts that will reject noisy epochs and retain clean epochs
+- Do the following for a set of possible candidate thresholds Ts
+- For each T:
+    - Split your data into K folds (=k equal parts) along the trial dimension
+        - Each of the k parts will be a "test" set once, while the others will be combined to be a "train" set
+    - Then for each fold k (consisting of train and test trials) do:
+        - apply threshold T to reject trials in the train set
+        - calculate the mean of the signal (for each sensor and timepoint) over the GOOD trials in the train set
+        - calculate the *median* of the signal (for each sensor and timepoint) over ALL trials in the test set
+        - compare both of these signals and calculate the error (frobenius norm of their difference)
+        - save the error
+    - Now we have k errors
+    - Form the mean error (over all k errors) associated with our current threshold T in microvolts
+    - Save the mapping of current T to its error
+
+- ... now we have a set of Ts all mapped to their specific error
+- the T with the lowest error is the best rejection threshold for a global rejection
+
+References
+----------
+.. [1] Mainak Jas, Denis Engemann, Yousra Bekhti, Federico Raimondo, and
+   Alexandre Gramfort. 2017. Autoreject: Automated artifact rejection for MEG
+   and EEG data. NeuroImage, 159, 417-429.
--- a/doc/explanation.rst
+++ b/doc/explanation.rst
@@ -1,5 +1,5 @@
-Explanations
-============
+Explanation
+===========
 
 This section of the documentation exists to illuminate how ``autoreject`` works.
 The primary source for understanding should be the original publication [1]_,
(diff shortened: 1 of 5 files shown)


jasmainak:
I can also share some powerpoints if you want to link them here.

My approach would be to make the explanation part also with data as it will make it more concrete + convincing. Basically show evoked (with spatial colors) for a couple of thresholds (low1, low2, optimal, high). Then show the RMSE for different thresholds with one train-test split. It should look jagged. Then average it across 5 folds to give a smoother picture. Next say that we can do it smarter/faster than grid search by showing how the gaussian process prediction and uncertainty estimates looks like (c.f. PR of @agramfort https://github.com/autoreject/autoreject/pull/129). Finally you can say that since it's faster we can now run this for all the channels and show the histogram of thresholds.

jasmainak:
although I realize that this taking you far from your GSoC work, so feel free to say no if you don't have the time / energy for such a thing.

sappelhoff:
> I can also share some powerpoints if you want to link them here.

that'd be a great addition - if you give me the links, I can weave them in

regarding your suggestions in https://github.com/autoreject/autoreject/pull/147#issuecomment-499197284 --> I really like these ideas and I think that it's the way to go ... but as you already pointed out: It'll lead me too far off my GSoC goal, so I'd rather keep this PR "a good small improvement" versus "a major great improvement"

So I'd streamline the "intuition" in `explanation.rst` and use a bit of math notation but then would be ready to have this merged. WDYT?

jasmainak:
okay, as long as you agree it's a good direction it's fine. I can push to this PR and let you review so we can merge in a couple of days :-) Last PR before the release I'd say

sappelhoff:
sounds great! If you manage the PR within the next 4 days, the release can be exactly one year after 0.1 ;-)

image

doc/explanation.rst
```diff
@@ -0,0 +1,47 @@
...
+The primary source for understanding should be the original publication [1]_,
+however the sections in this guide can make the content of that primary source
+more graspable.
+
+
+Intuition on how the - *autoreject global* - algorithm works
+------------------------------------------------------------
+- Given some MEEG data :math:`X` with the dimensions
+  :math:`trials(=epochs) \times sensors \times timepoints`
+- We want to find a threshold :math:`\phi` in :math:`\mu V` that will reject
```
jasmainak: I would use the same notation as in the paper, that is calling it "tau"

doc/explanation.rst
```diff
@@ -0,0 +1,47 @@
...
+  noisy epochs and retain clean epochs
+- Do the following for a set of possible candidate thresholds: :math:`\Phi`
+- For each :math:`\phi_i \in \Phi` :
+    - Split your data :math:`X` into :math:`K` folds (=:math:`K` equal parts)
+      along the trial dimension
+        - Each of the :math:`K` parts will be a "test" set once, while the
+          remaining :math:`K-1` parts will be combined to be the corresponding
+           "train" set (see `k-fold crossvalidation <https://en.wikipedia.org/wiki/Cross-validation_(statistics)#k-fold_cross-validation>`_)
```
jasmainak: this part is not rendered correctly

doc/explanation.rst
```diff
@@ -0,0 +1,47 @@
+Explanations
```
jasmainak: Explanations -> Explanation

doc/explanation.rst
sappelhoff: Should be fixed through the help of [rst docs](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#bullet-lists)

sappelhoff:
rendered version with all comments addressed: https://616-59485140-gh.circle-artifacts.com/0/html/explanation.html

cc @jasmainak

jasmainak:
Merged. Thanks @sappelhoff !

ready for release now.

PR #207, opened 2021-04-26
[MRG, BUG] fix long computation time in example

alexrockhill:
Related to https://github.com/autoreject/autoreject/pull/205

This example takes ~70 minutes to run for some reason. This fixes that by picking a subset of channels but leaving the one that really makes the example.

### DIFF
--- a/examples/plot_auto_repair.py
+++ b/examples/plot_auto_repair.py
@@ -11,7 +11,7 @@ repair epochs.
 #         Denis A. Engemann <denis.engemann@gmail.com>
 # License: BSD (3-clause)
 
-# %%
+###############################################################################
 # Let us first define the parameters. `n_interpolates` are the :math:`\rho`
 # values that we would like :mod:`autoreject` to try and `consensus_percs`
 # are the :math:`\kappa` values that :mod:`autoreject` will try (see the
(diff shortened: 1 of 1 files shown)


alexrockhill:
70 minutes down to 40 seconds with all the same content, looks good to go by me https://774-59485140-gh.circle-artifacts.com/0/html/auto_examples/plot_auto_repair.html#sphx-glr-auto-examples-plot-auto-repair-py

jasmainak:
I think we should fix the root of the problem, not change the example. This example reproduces Figure 1 of the paper. It seems to be a config issue with CircleCI

alexrockhill:
Perhaps a RAM issue as mentioned in another comment, making it use swap. Wouldn't it be easier to leave it to execute in 30 seconds rather then 10 minutes and say that the paper could be reproduced without the decimation?

agramfort:
did you run memory_profiler locally to see how much RAM the example uses?
CircleCI has 4GB of RAM AFAIK.

could it be a regression in mne-python?

>

jasmainak:
no memory issues I think @agramfort 

image

jasmainak:
I enabled this option:

image

it should speed things up quite a bit, specially when folks are pushing multiple commits one after the other.

jasmainak:
@alexrockhill I'm going to close this issue for now, I think it was some temporary slowness you noticed because of pushing multiple times. We can reopen if the issue persists.

alexrockhill:
Sounds good, if it's working fast it's working fast, it was just really slow on many CI runs so the servers could have been running slow or something but they definitely weren't all simultaneous so that doesn't seem like the most likely cause of it being slow.

PR #209, opened 2021-04-26
[MRG][ENH] Make reject log plot prettier

alexrockhill:
Related to #205.

Change the coloring to be more intuitive (green=good, red=bad) and add red boxes around rejected epochs.

### DIFF
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -1217,6 +1217,7 @@ class RejectLog(object):
             it shows every 5 entries. If int, show every show_names entries.
         show : bool
             If True, display the figure immediately.
+
         Returns
         -------
         figure : Instance of matplotlib.figure.Figure
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -1224,8 +1224,11 @@ class RejectLog(object):
         figure, ax = plt.subplots(figsize=(12, 6))
         ax.grid(False)
 
+        labels = self.labels.copy()
+        labels[labels > 0] = 1 / labels[labels > 0]  # move interp to 0.5
+
         if orientation == 'horizontal':
-            img = ax.imshow(self.labels.T / 2, cmap='RdYlGn_r',
+            img = ax.imshow(labels, cmap='RdYlGn_r',
                             vmin=0, vmax=1, interpolation='nearest')
             ax.set_xlabel('Epochs')
             ax.set_ylabel('Channels')
@@ -1238,7 +1241,7 @@ class RejectLog(object):
                     (idx - 0.5, -0.5), 1, len(self.ch_names), linewidth=1,
                     edgecolor='r', facecolor='none'))
         elif orientation == 'vertical':
-            img = ax.imshow(self.labels / 2, cmap='RdYlGn_r',
+            img = ax.imshow(labels, cmap='RdYlGn_r',
                             vmin=0, vmax=1, interpolation='nearest')
             ax.set_xlabel('Channels')
             ax.set_ylabel('Epochs')
@@ -1256,9 +1259,9 @@ class RejectLog(object):
             raise ValueError(msg)
 
         # add legend
-        handles = [patches.Patch(color=img.cmap(img.norm(i / 2)), label=label)
+        handles = [patches.Patch(color=img.cmap(img.norm(i)), label=label)
                    for i, label in
-                   {0: 'good', 1: 'bad', 2: 'interpolated'}.items()]
+                   {0: 'good', 1: 'bad', 0.5: 'interpolated'}.items()]
         ax.legend(handles=handles, bbox_to_anchor=(1.05, 1), loc=2,
                   borderaxespad=0.)
 
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -1226,7 +1226,7 @@ class RejectLog(object):
 
         image = self.labels.copy()
         image[image == 2] = 0.65  # move interp to 0.5
-        image[image == 0] = np.nan  # no color for good channels
+        image[image == 0] = 0.15  # no color for good channels
 
         if orientation == 'horizontal':
             img = ax.imshow(image.T, cmap='RdYlGn_r',
@@ -1262,7 +1262,7 @@ class RejectLog(object):
         # add legend
         handles = [patches.Patch(color=img.cmap(img.norm(i)), label=label)
                    for i, label in
-                   {np.nan: 'good', 1: 'bad', 0.65: 'interpolated'}.items()]
+                   {0.15: 'good', 1: 'bad', 0.65: 'interpolated'}.items()]
         ax.legend(handles=handles, bbox_to_anchor=(1.05, 1), loc=2,
                   borderaxespad=0.)
 
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -1213,8 +1213,9 @@ class RejectLog(object):
             If `'horizontal'`, will plot epochs on x-axis and sensors
             on y-axis.
         show_names : 'auto' | int
-            If 'auto', show all channel names if fewer than 25 entries. Otherwise
-            it shows every 5 entries. If int, show every show_names entries.
+            If 'auto', show all channel names if fewer than 25 entries.
+            Otherwise it shows every 5 entries. If int, show every
+            show_names entries.
         show : bool
             If True, display the figure immediately.
 
@@ -1254,9 +1255,9 @@ class RejectLog(object):
 
             # add legend
             handles = [patches.Patch(color=img.cmap(img.norm(i)), label=label)
-                        for i, label in legend_label.items()]
+                       for i, label in legend_label.items()]
             ax.legend(handles=handles, bbox_to_anchor=(3.5, 0.5), ncol=1,
-                    borderaxespad=0.)
+                      borderaxespad=0.)
 
         elif orientation == 'vertical':
             img = ax.imshow(image, cmap=cmap,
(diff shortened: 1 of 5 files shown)


jasmainak:
On the website I have:

image

on this PR, I have:

image

I think I prefer the former ...

alexrockhill:
> On the website I have:
> 
> image
> 
> on this PR, I have:
> 
> image
> 
> I think I prefer the former ...

Look how it turns out on the full PR, much better. Having too many channels can be solved by `fig.gca().set_xticks(range(0, len(epochs.ch_names), 10)` for instance, that's much better than the arbitrary amount used in the first plot.

alexrockhill:
See this artifact for how it would look with the example changed to match https://771-59485140-gh.circle-artifacts.com/0/html/auto_examples/plot_example_workflow.html#sphx-glr-auto-examples-plot-example-workflow-py

jasmainak:
@alexrockhill better! But now the legend does not match the colors:

image

also "interpolated" is very hard to see

jasmainak:
I would make the green even lighter ... the focus should be on the bad and interpolated channels

dengemann:
I like the idea. Perhaps legend labels can be drawn smaller and arranged horizontally to leave more room for actual data (information / ink ratio).

jasmainak:
Looks good now! +1 for MRG after moving the legend to the bottom as @dengemann suggested

jasmainak:
@alexrockhill do you want to update `whats_new.rst` ?

autoreject/autoreject.py
```diff
@@ -1203,8 +1203,8 @@ def __init__(self, bad_epochs, labels, ch_names):
         assert len(bad_epochs) == labels.shape[0]
         assert len(ch_names) == labels.shape[1]
 
-    def plot(self, orientation='vertical', show=True):
-        """Plot.
+    def plot(self, orientation='horizontal', show=True):
```
jasmainak, line 1206: what's the reason for changing this default?

hoechenberger:
IMHO the green is superfluous and only makes it more difficult to actually spot the bad and interpolated segments. Also from an accessibility perspective, choosing red and green is probably not the best idea.

I’d therefore propose to simply drop the green

alexrockhill:
I liked the dark green because it really made the red and yellow pop but no color for good is fine too.

jasmainak:
But now the problem is that, we don't see which channels are part of the `picks`. Earlier we showed all channels and channels that were not part of the picks were shown in "no color" (gray)

image

hoechenberger:
Oh. Ok. Well I don't insist on the change I proposed. Just wanted to point out that there might be an accessibility problem there…

agramfort:
I find the interpolated not enough visible 😅😌😌

>

hoechenberger:
> I find the interpolated not enough visible 😅😌😌

Same!

jasmainak:
let me do a pass

jasmainak:
I just pushed a commit. How do folks like this:

image

I'm going to add a `show_names` argument a la [`evoked.plot_image`](https://mne.tools/stable/generated/mne.Evoked.html#mne.Evoked.plot_image) so folks can control how dense the tick labels are.

@alexrockhill can you try on your datasets and share if they look fine on them?

agramfort:
still a bit to red/green. See test on colorblind vision

Screenshot 2021-05-04 at 15 10 40
Screenshot 2021-05-04 at 15 10 32

simulated at https://www.color-blindness.com/coblis-color-blindness-simulator/

jasmainak:
colortest

image

it seems to survive colorblind tests if I make it blue. Would this be okay?

jasmainak:
tests will pass after #211 is fixed

alexrockhill:
https://www.nature.com/articles/nmeth.1618

Maybe vermillion=bad, yellow=interpolated, bluish green=good. That's from looking at figure 2 and picking from there.

agramfort:
let's go for it.

I think it's a real progress anyway.

sorry to nitpick...

>

sappelhoff:
> tests will pass after #211 is fixed

just merged it - @alexrockhill you can rebase :)

jasmainak:
@alexrockhill the rebase lost my commits? :( I'm going to rebase with what was latest ...

jasmainak:
okay so the Nature color combo doesn't really work @alexrockhill (aside: the paper says RGB colors are from 1-255, it should be 0-255)

image

it's too crowded. I am happy with the current selection as it survived the [colorblind tests](https://github.com/autoreject/autoreject/pull/209#issuecomment-831971883). Good to merge from my end

jasmainak:
I also tried making the green lighter, that doesn't work either:

image

Blue is better than yellow IMO

jasmainak:
@hoechenberger any clue what the CircleCI error is about? It's complaining about the `openneuro` package

hoechenberger:
It appears OpenNeuro has finally deisabked the previously deprecated API that `openneuro-py` is using. I have just started working on updating it, but I won't finish this today. Should have it ready by the end of the weekend though!

jasmainak:
alright! I can wait :)

hoechenberger:
@jasmainak OpenNeuro PR: https://github.com/hoechenberger/openneuro-py/pull/38

hoechenberger:
@jasmainak Using `openneuro-py:main` should solve the issue for you.
If it does work, I'll try to make a release soon.

jasmainak:
Thanks a lot @hoechenberger . Fingers crossed the CIs come green!

hoechenberger:
@jasmainak You don't want the URL in setup.py, this should go to a concrete environment / requirements definition :)

hoechenberger:
I'll make a release shortly so you don't have to worry about this.

hoechenberger:
@jasmainak `openneuro-py 2021.3` is on PyPI. You should probably add this version constraint to `setup.py`

jasmainak:
Thanks for the release! Actually, it still wasn't working when I used the URL with the development version. Let's see if it works with the release version ...

hoechenberger:
@jasmainak This is a server timeout, that should be unrelated to `openneuro-py` itself. I tried this one locally and I can replicate the issue. I will try again in a few hours and if it persists, I'm afraid we'll have to get in touch with OpenNeuro folks

hoechenberger:
It appears we're hitting an implicit and new concurrency limit.

A workaround is to limit the number of parallel downloads to 1:

in `examples/plot_visualize_bad_epochs.py`, change 

```python
openneuro.download(dataset=dataset, target_dir=target_dir,
                   include=[f'sub-{subject_id}/ses-meg/'])
```

to

```python
openneuro.download(dataset=dataset, target_dir=target_dir,
                   include=[f'sub-{subject_id}/ses-meg/'],
                   max_concurrent_downloads=1)
```

hoechenberger:
@jasmainak NB for the above to work, you'll need to use `openneuro-py:main` again, sorry :(

hoechenberger:
@jasmainak I've released `openneuro-py` 2021.4, so you don't have to use `main`.

hoechenberger:
@jasmainak I've published `openneuro-py` 2021.5 that should be even more robust in dealing with OpenNeuro's flaky servers.

jasmainak:
@hoechenberger apologies for the slow responses here. Autoreject maintenance is happening mostly during my free time. I've pushed an update to use the latest `openneuro`. Fingers crossed!

hoechenberger:
Aww come on… Still failing. I'll have a look.

hoechenberger:
It just says that the process got killed. Could that be a memory issue? `openneuro-py` is supposed to stream the data, writing chunks to disk as they are retrieved, so it won't fill up memory. But I've never profiled actual memory usage. I wonder if I need to do that now … WDYT, @agramfort?

hoechenberger:
But actually, we're performing the exact same download in `openneuro-py`'s CI and it works without issues there:

https://github.com/hoechenberger/openneuro-py/blob/96c803dca57017b35d8b46ea2c9d2fbc52c35901/openneuro/tests/test_download.py#L9-L10

Then again this is run on GH Actions and not Circle

hoechenberger:
@jasmainak If you're okay with temporarily granting me the required access rights on this repo, I can try to get to the bottom of this myself, without bothering you :)

jasmainak:
@hoechenberger I just gave you write access!

jasmainak:
you can take over this PR

hoechenberger:
Thanks! Will look into this tomorrow!

hoechenberger:
It seems I cannot control Circle CI (I might need to SSH into the CI system), I'm not sure which permissions would be required for this..?

jasmainak:
I bumped your privilege to "Maintain" ... hope that does the trick :)

hoechenberger:
Whoops.

Screen Shot 2021-05-23 at 20 37 42

hoechenberger:
Did some small optimizations:

Screen Shot 2021-05-23 at 22 01 17

Reduced RAM consumption by 2 orders of magnitude.

Will push this to `openneuro-py` tomorrow.

jasmainak:
wow!! that's awesome. Autoreject is pushing the boundaries of openneuro-py ;-)

setup.py
```diff
@@ -77,7 +77,7 @@
                   'numpydoc',
                   'cython',
                   'pillow',
-                  'openneuro-py'
+                  'openneuro-py >= 2021.5'
```
hoechenberger, line 80:
```suggestion
                  'openneuro-py >= 2021.7'
```

hoechenberger:
I didn't squash this one before merging, ouch. How should we go about this?

jasmainak:
all good. I tend not to squash in younger projects so everyone gets due credit. But disabled the merge commit option now so as to keep linear history

jasmainak:
Thanks for the fixes!

PR #225, opened 2021-08-06
Fix #224 - Allow passing a valid string to picks.

### LINKED ISSUE #224: IndexError when using Ransac.fit()
I ran into an issue with my data which I could reproduce using a dummy example with MNE sample data:

```python
from pathlib import Path

import mne
from autoreject import Ransac

sample_data_dir = Path(mne.datasets.sample.data_path())
fname = sample_data_dir / 'MEG' / 'sample' / 'sample_audvis_raw.fif'

raw = mne.io.read_raw_fif(fname)
raw.del_proj()
raw.pick(('eeg', 'stim'))

raw.set_eeg_reference(ref_channels='average', ch_type='eeg', projection=True)

events = mne.find_events(raw, stim_channel='STI 001')
epochs = mne.Epochs(
    raw, events, event_id=5, picks='eeg', tmin=-0.2, tmax=0.8,
    reject=None, verbose=False, detrend=None, proj=True,
    baseline=(None, 0), preload=True)

ransac = Ransac(verbose=False, picks='eeg', n_jobs=1)
ransac.fit(epochs)
```

Error: 

> IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices

Traceback:

> File "/var/folders/b1/kx2x00x95ls80zzkct4647sh0000gn/T/ipykernel_4007/3370373497.py", line 22, in <module>
>     ransac.fit(epochs)
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/autoreject/ransac.py", line 187, in fit
>     _check_data(epochs, picks=self.picks,
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/autoreject/utils.py", line 40, in _check_data
>     picked_info = mne.io.pick.pick_info(epochs.info, picks)
> 
>   File "<decorator-gen-8>", line 24, in pick_info
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/io/pick.py", line 540, in pick_info
>     n_unique = len(np.unique(np.arange(len(info['ch_names']))[sel]))
> 
> IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices

I'm using the latest version of `MNE` (0.23) and `autoreject` (0.2.1).

mscheltienne:
Fixes issue #224.

Use MNE `_picks_to_idx()` function to handle picks. Supports both strings and array of indices.
Added test for a string argument.

Is there a changelog to update? Anything else which I might add?

### DIFF
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -615,16 +615,24 @@ class _AutoReject(BaseAutoReject):
         ----------
         epochs : instance of mne.Epochs
             The epochs from which to get the drop logs.
-        picks : np.ndarray, shape(n_channels, ) | list | None
-            The channel indices to be used. If None, the .picks attribute
-            will be used.
+        picks : str | list | slice | None
+            Channels to include. Slices and lists of integers will be
+            interpreted as channel indices. In lists, channel *type* strings
+            (e.g., ``['meg', 'eeg']``) will pick channels of those types,
+            channel *name* strings (e.g., ``['MEG0111', 'MEG2623']`` will pick
+            the given channels. Can also be the string values ``'all'`` to pick
+            all channels, or ``'data'`` to pick data channels. None (default)
+            will use the .picks attribute. Note that channels in
+            ``info['bads']`` *will be included* if their names or indices are
+            explicitly provided.
 
         Returns
         -------
         reject_log : instance of autoreject.RejectLog
             The rejection log.
         """
-        picks = (self.picks_ if picks is None else picks)
+        picks = (self.picks_ if picks is None else
+                 _handle_picks(epochs.info, picks))
         picks_by_type = _get_picks_by_type(picks=picks, info=epochs.info)
         assert len(picks_by_type) == 1
         ch_type, this_picks = picks_by_type[0]
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -471,9 +471,14 @@ class _AutoReject(BaseAutoReject):
     thresh_func : callable | None
         Function which returns the channel-level thresholds. If None,
         defaults to :func:`autoreject.compute_thresholds`.
-    picks : ndarray, shape(n_channels,) | None
-        The channels to be considered for autoreject. If None, defaults
-        to data channels {'meg', 'eeg'}.
+    picks : str | list | slice | None
+        Channels to include. Slices and lists of integers will be interpreted
+        as channel indices. In lists, channel *type* strings (e.g.,
+        ``['meg', 'eeg']``) will pick channels of those types, channel *name*
+        strings (e.g., ``['MEG0111', 'MEG2623']`` will pick the given channels.
+        Can also be the string values ``'all'`` to pick all channels, or
+        ``'data'`` to pick data channels. None (default) will pick data
+        channels {'meg', 'eeg'}. Note that channels in ``info['bads']``
     verbose : 'tqdm', 'tqdm_notebook', 'progressbar' or False
         The verbosity of progress messages.
         If `'progressbar'`, use `mne.utils.ProgressBar`.
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -375,9 +375,15 @@ def compute_thresholds(epochs, method='bayesian_optimization',
         'bayesian_optimization' or 'random_search'
     random_state : int seed, RandomState instance, or None (default)
         The seed of the pseudo random number generator to use
-    picks : ndarray, shape(n_channels,) | None
-        The channels to be considered for autoreject. If None, defaults
-        to data channels {'meg', 'eeg'}.
+    picks : str | list | slice | None
+        Channels to include. Slices and lists of integers will be interpreted
+        as channel indices. In lists, channel *type* strings (e.g.,
+        ``['meg', 'eeg']``) will pick channels of those types, channel *name*
+        strings (e.g., ``['MEG0111', 'MEG2623']`` will pick the given channels.
+        Can also be the string values ``'all'`` to pick all channels, or
+        ``'data'`` to pick data channels. None (default) will pick data
+        channels {'meg', 'eeg'}. Note that channels in ``info['bads']`` *will
+        be included* if their names or indices are explicitly provided.
     augment : boolean
         Whether to augment the data or not. By default it is True, but
         set it to False, if the channel locations are not available.
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -852,11 +852,17 @@ class AutoReject(object):
         This is :math:`\\rho`. If None, defaults to np.array([1, 4, 32])
     cv : a scikit-learn cross-validation object
         Defaults to cv=10
-    picks : ndarray, shape(n_channels) | None
-        The channels to be considered for autoreject. If None, defaults
-        to data channels {'meg', 'eeg'}, which will lead fitting and combining
-        autoreject solutions across these channel types. Note that, if picks is
-        None, autoreject ignores channels marked bad in epochs.info['bads'].
+    picks : str | list | slice | None
+        Channels to include. Slices and lists of integers will be interpreted
+        as channel indices. In lists, channel *type* strings (e.g.,
+        ``['meg', 'eeg']``) will pick channels of those types, channel *name*
+        strings (e.g., ``['MEG0111', 'MEG2623']`` will pick the given channels.
+        Can also be the string values ``'all'`` to pick all channels, or
+        ``'data'`` to pick data channels. None (default) will pick data
+        channels {'meg', 'eeg'}, which will lead fitting and combining
+        autoreject solutions across these channel types. Note that channels in
+        ``info['bads']`` *will be included* if their names or indices are
+        explicitly provided.
     thresh_method : str
         'bayesian_optimization' or 'random_search'
     n_jobs : int
--- a/autoreject/autoreject.py
+++ b/autoreject/autoreject.py
@@ -947,7 +947,7 @@ class AutoReject(object):
         self : instance of AutoReject
             The instance.
         """
-        self.picks_ = _handle_picks(picks=self.picks, info=epochs.info)
+        self.picks_ = _handle_picks(info=epochs.info, picks=self.picks)
         _check_data(epochs, picks=self.picks_, verbose=self.verbose)
         self.cv_ = self.cv
         if isinstance(self.cv_, int):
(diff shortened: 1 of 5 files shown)


autoreject/autoreject.py
```diff
@@ -947,7 +947,7 @@ def fit(self, epochs):
         self : instance of AutoReject
             The instance.
         """
-        self.picks_ = _handle_picks(picks=self.picks, info=epochs.info)
+        self.picks_ = _handle_picks(info=epochs.info, picks=self.picks)
```
jasmainak, line 950: can you revert changes unrelated to the PR?

autoreject/tests/test_ransac.py
```diff
@@ -36,16 +36,23 @@ def test_ransac():
     ransac = Ransac(picks=picks, random_state=np.random.RandomState(42))
     epochs_clean = ransac.fit_transform(epochs)
     assert len(epochs_clean) == len(epochs)
+
+    # Pass string instead of array of idx
+    picks = 'eeg'
+    ransac = Ransac(picks=picks, random_state=np.random.RandomState(42))
+    epochs_clean = ransac.fit_transform(epochs)
+    assert len(epochs_clean) == len(epochs)
```
jasmainak, line 44: this is what one would call a smoke test. It's fine for a first pass but it doesn't explicitly test the `picks`. Can you check `ransac.picks` and assert that the channels are indeed all EEG channels (perhaps checked against output of `mne.pick_types`). Also, in order to keep the test fast you could `ransac.fit(epochs[:2])` or something like that

jasmainak:
You can update `whats_new.rst`. Thanks for the fix!

autoreject/autoreject.py
mscheltienne: Sure I can, since I changed `_handle_picks`, I went through the files where `_handle_picks` was used; I stumble upon this entry where the arguments are provided in the wrong order, sure with keyword.. but still. It just makes the code a little bit more messy for future improvements in my opinion. Why not improve it together with the fix for this function?

autoreject/tests/test_ransac.py
mscheltienne: Yes ok, this test was mostly just to check that it was not crashing. I'll add an assertion on a matching condition.

autoreject/autoreject.py
jasmainak: I see. The rationale is to keep the diff as clean and small as possible to make the reviewers job easy. For example, I spent a good 30 seconds here wondering if something was broken here and if so, why there was no test for the `Autoreject` class. Until I realized that the two lines were the same ...

autoreject/tests/test_ransac.py
mscheltienne: Lucky enough I used `picks='eeg'`. If you set this test with `picks='mag'` and assert `ransac.picks` against `mne.pick_types(epochs.info, meg='mag', eeg=False, stim=False, eog=False, exclude=[])`, it matches. However, with EEG, one of the channels is dropped and ``ransac.picks` is missing one channel. Looking into that.

autoreject/autoreject.py
mscheltienne: Ok, I can understand this rationale. However, it also means that to improve existing code, to develop new features, you also need a good 30 seconds to understand why it is written this way. And this is even worst for people which are not familiar with the code of a given repo. In my opinion, this is why there are standards like PEP8 and static analysis tools to improve code readability and help X understand the code written by Y. 
If anything, this confusion was my mistake for not adding to the description of the PR an explanation on this change, even if it was minor.
If you confirm that you want this change revert, I will; but now that you did spend those 30 seconds realizing it was the same code.. wouldn't it be better to keep it as it is in the end an improvement on the former code which might prevent future confusion?

autoreject/tests/test_ransac.py
mscheltienne: Ok, I'm just a bit of an idiot. One channel was marked in `info['bads']`. This brings me to another problem: the default behavior of `_handle_picks(info, picks)` when `picks=None` was to exclude bads. Thus, on the change I made I also excluded bads. However in the test, and in my mind, it makes a lot more sense for autoreject, the `picks` are not excluding any channels. @jasmainak So is this exclusion of marked bad channels intended or is this a mistake? According to your reply, I will update accordingly `_handle_picks()` and the test with either an assertion against `picks = mne.pick_types(epochs.info, meg=False, eeg=True, stim=False, eog=False, exclude=[])` or `exclude='bads'`.

autoreject/autoreject.py
mscheltienne: Depending on this inclusion or exclusion of bad channels, I will update the different docstring with something similar to MNE.

autoreject/tests/test_ransac.py
mscheltienne: Depending on this inclusion or exclusion of bad channels, I will update the different docstrings with something similar to MNE.

jasmainak: I think that bad channels that were marked by the user should be ignored by Autoreject and Ransac.

autoreject/autoreject.py
jasmainak: haha let's keep it as is :) General principle is to make cosmetic changes in separate PRs when possible.

mscheltienne: I'll keep it in mind for future contributions ;)

autoreject/tests/test_ransac.py
mscheltienne: Ok, I'll make the update accordingly.

mscheltienne:
@jasmainak One last point on the docstring. I wanted to update the docstring for `picks` with something along the line of:

> picks : str | list | slice | None
>         Channels to include. Slices and lists of integers will be interpreted
>         as channel indices. In lists, channel *type* strings (e.g.,
>         ``['meg', 'eeg']``) will pick channels of those types, channel *name*
>         strings (e.g., ``['MEG0111', 'MEG2623']`` will pick the given channels.
>         Can also be the string values ``'all'`` to pick all channels, or
>         ``'data'`` to pick data channels. None (default) will pick data
>         channels {'meg', 'eeg'}. Note that channels in ``info['bads']`` *will
>         be included* if their names or indices are explicitly provided.

Which is the docstirng used by MNE (mostly) adapted for autoreject. Now in MNE, to add a piece of code to the documentation, I believe the syntax is ``code``; but it doesn't seem to be the case in autoreject. I see a lot of single *whatever this quote symbol is named* instead of double before/after. e.g.

image

Which results in this documentation:

image

Is this intended or should all those single *quote symbol name* be doubled for a better rendering of the documentation?

mscheltienne:
And one more point, would you like me to disable the functionality `picks='all'`? An error is raised anyway if multiple channel types are provided.

jasmainak:
> Is this intended or should all those single quote symbol ` be doubled for a better rendering of the documentation?

this is unintended. You can fix them in this PR if you like

> And one more point, would you like me to disable the functionality picks='all'? An error is raised anyway if multiple channel types are provided. But it could still be used by users which have only one datatype in an instance.

humm ... it depends where this is being used. You can do multiple channel types in autoreject, it will take the union of the epochs from each channel type. I'm not sure if it's the case for Ransac though

mscheltienne:
Alright, I will change the docstring for `picks`, with the base written above; and I will adapt it for each individual function.

mscheltienne:
So that will be it for this PR which should cover all the changes related to `_handle_picks()`. Have a careful look at the different docstring. I did split each update into a different commit for convenience. 

The docstring is not always the same. I re-used (partially) the previously existing docstring. 
If the function does not take in input multiple data types, I removed the part describing list of channel types `['eeg', 'meg']`. 
For function working only on EEG, I changed the example from MEG to EEG.

As I am not familiar with (most) of those functions, have a careful look as I might have missed something, and a specific docstring might be improved further to better describe its behavior. 

You'll notice that I put a warning mark in one of the commit titles. The method `AutoReject.get_reject_log()` (line 1042) takes `picks` as argument; but does not use it. That combine with the comment `# XXX gut feeling that there is a bad condition that we miss`; I guess it might be related?

I also added in 2 places a call to `_handle_picks()`.

For the quotation marking in the documentation, I'll make a different PR as this is unrelated to `_handle_picks()` and to `picks` in general.

autoreject/tests/test_ransac.py
```diff
@@ -32,20 +32,29 @@ def test_ransac():
...
                            eog=False, exclude=[])
-
     ransac = Ransac(picks=picks, random_state=np.random.RandomState(42))
     epochs_clean = ransac.fit_transform(epochs)
     assert len(epochs_clean) == len(epochs)
+
+    # Pass string instead of array of idx
+    picks = 'eeg'
+    ransac = Ransac(picks=picks, random_state=np.random.RandomState(42))
+    epochs_clean = ransac.fit_transform(epochs[:2])
```
jasmainak, line 42: you can even just do:

```suggestion
    epochs = ransac.fit(epochs[:2])
```

to make it even faster. We really just want to test the `picks`. L43 is probably tested elsewhere

autoreject/tests/test_ransac.py
```diff
@@ -32,20 +32,29 @@ def test_ransac():
...
     picks = mne.pick_types(epochs.info, meg='mag', eeg=False, stim=False,
                            eog=False, exclude=[])
-
     ransac = Ransac(picks=picks, random_state=np.random.RandomState(42))
     epochs_clean = ransac.fit_transform(epochs)
     assert len(epochs_clean) == len(epochs)
+
+    # Pass string instead of array of idx
+    picks = 'eeg'
+    ransac = Ransac(picks=picks, random_state=np.random.RandomState(42))
```
jasmainak, line 41:
```suggestion
    ransac = Ransac(picks=picks)
```

better to leave out `random_state` since it shouldn't affect picks selection

jasmainak (approved):
Looks good to me otherwise. @dengemann do you want to take a look as you implemented the picks ?

autoreject/tests/test_ransac.py
mscheltienne: Sure. The test right above is very slow, maybe you could push a fix where only 2 epochs are used instead of the 4?

jasmainak:
Merging to keep the ball rolling. Thanks @mscheltienne for the contribution. Looking forward to the next PR :)

PR #227, opened 2021-08-12
allow verbose to be 'tqdm_notebook' and 'progressbar'

pstetz:
It looks like whatever `verbose` option was chosen is overwritten [here](https://github.com/autoreject/autoreject/blob/c3f5a8186ed15fd7e16a4f44d36a3f22390ee2c4/autoreject/utils.py#L178). I deleted this section so that the user supplied option is chosen.

Also changed `tqdm_notebook` to `tqdm.notebook` because the former is deprecated

### DIFF
--- a/autoreject/utils.py
+++ b/autoreject/utils.py
@@ -173,24 +173,19 @@ def _pbar(iterable, desc, leave=True, position=None, verbose='progressbar'):
         raise ValueError('verbose must be one of {progressbar,'
                          'tqdm, tqdm_notebook, False}. Got %s' % verbose)
 
-    try:
-        from tqdm import tqdm
-        verbose = 'tqdm'
-    except ImportError:
-        pass
-
     if verbose == 'progressbar':
         from mne.utils import ProgressBar
         pbar = ProgressBar(iterable, mesg=desc)
     # XXX: remove the tqdm option after a few releases of MNE since it
     # natively supported by the MNE progressbar
     elif verbose == 'tqdm':
+        from tqdm import tqdm
         pbar = tqdm(iterable, desc=desc, leave=leave, position=position,
                     dynamic_ncols=True)
     elif verbose == 'tqdm_notebook':
-        from tqdm import tqdm_notebook
-        pbar = tqdm_notebook(iterable, desc=desc, leave=leave,
-                             position=position, dynamic_ncols=True)
+        from tqdm.notebook import tqdm
+        pbar = tqdm(iterable, desc=desc, leave=leave,
+                position=position, dynamic_ncols=True)
     elif verbose is False:
         pbar = iterable
     return pbar


sappelhoff (approved):

autoreject/utils.py
```diff
@@ -173,24 +173,19 @@ def _pbar(iterable, desc, leave=True, position=None, verbose='progressbar'):
...
+        from tqdm import tqdm
         pbar = tqdm(iterable, desc=desc, leave=leave, position=position,
                     dynamic_ncols=True)
     elif verbose == 'tqdm_notebook':
-        from tqdm import tqdm_notebook
-        pbar = tqdm_notebook(iterable, desc=desc, leave=leave,
-                             position=position, dynamic_ncols=True)
+        from tqdm.notebook import tqdm
+        pbar = tqdm(iterable, desc=desc, leave=leave,
+                position=position, dynamic_ncols=True)
```
sappelhoff, line 188:
```suggestion
        pbar = tqdm(iterable, desc=desc, leave=leave,
                    position=position, dynamic_ncols=True)
```

jasmainak:
Can you check that autoreject can be run without a tqdm dependency after your changes? I need to dig why I introduced that try/pass logic ... might be a bug of course

pstetz:
Hi @jasmainak, [tqdm is a dependency of mne](https://github.com/mne-tools/mne-python/blob/v0.23.0/requirements.txt#L32) so I think it would only be an issue for older versions. Also my reasoning is that the import statement will break only if the user supplies the "tqdm" or "tqdm_notebook" option which would be weird if they didn't have it installed

This [PR](https://github.com/mne-tools/mne-python/pull/7155) from MNE and this [PR](https://github.com/autoreject/autoreject/commit/9902dafeff1f7be34cc891b411b148016dcc951c) from autoreject seem relevant. Not sure if they will jog your memory

I think it's coded this way because the mne progressbar is the default option and doesn't seem to work (see below when I ran autoreject with my modified PR). Because there's something funny with the "progressbar" option, I think you forced "tqdm" when that's available
Screen Shot 2021-08-13 at 11 55 38 AM

I can modify the code so that all `verbose` options default to "tqdm" instead of "progressbar"?

jasmainak:
@pstetz apologies for the delayed response. I was away from computer for a few days

Thanks for taking the initiative to dig out the relevant PRs. Some time ago MNE moved to tqdm progressbars but they vendor it in externals as well. So it's not technically a dependency even if it's included in the environment.yml. We could use `mne.externals.tqdm` and/or `mne.ProgressBar` that would avoid adding the dependency? I'm okay deprecating two options out of "tqdm_notebook", "tqdm" and "progressbar" options since they are all flavors of `tqdm`. 

I believe `tqdm` also has an "auto" option now to switch between `tqdm` and `tqdm_notebook`, which MNE probably uses? I'm not so sure because the output you post above seems to suggest that MNE is not importing the right version of `tqdm` in the notebook automatically. Anyhow, if this were possible, it would obviate the need for `tqdm_notebook` as a separate option.

agramfort:
+1 to use mne Progressbar that uses tqdm internally

>

pstetz:
Good point! Just looked into MNE's progressbar and it makes Auotreject's implementation cleaner. Are you okay if `verbose` is boolean now like below?

```python3
def _pbar(iterable, desc, leave=True, position=None, verbose=True):
    if verbose:
        from mne.utils.progressbar import ProgressBar                                                                                                                                                                                                                                                                                                                     
        pbar = ProgressBar(iterable, mesg=desc)
    else:
        pbar = iterable
    return pbar
```

The above works for me

jasmainak:
yeah, but I guess we should check if:

1) this works in Jupyter notebooks
2) it doesn't break older scripts. Likely we should still allow the older options at least for one release cycle with a deprecation warning (hopefully outside the loops so it's printed only once ...)

pstetz:
1.) It works in both Jupyter notebooks and from Terminal
2.) What do you mean about "older options"? I thought the only options are tqdm or MNE's Progressbar (which internally is tqdm)

If a user provides `verbose=tqdm` or `verbose=tqdm_notebook`, it will be interpreted as a True boolean and the progressbar will show.

I can update autoreject's docstring's so that the "tqdm", "tqdm_notebook", and "progressbar" options aren't mentioned

jasmainak:
> If a user provides verbose=tqdm or verbose=tqdm_notebook, it will be interpreted as a True boolean and the progressbar will show.

correct, but you shouldn't allow a random string like 'blah' to be interpreted as `verbose=True`. Allow only `tqdm`, `tqdm_notebook` so user scripts are not broken. These are what I call "older options"

> I can update autoreject's docstring's so that the "tqdm", "tqdm_notebook", and "progressbar" options aren't mentioned

That sounds fantastic

pstetz:
Perfect, thank you for the clarification @jasmainak. Hopefully, with the latest commits all concerns are addressed, but let me know if you have other thoughts!

jasmainak:
you have flake8 issues @pstetz

autoreject/utils.py
```diff
@@ -165,33 +165,20 @@ def _pprint(params, offset=0, printer=repr):
...
-    # natively supported by the MNE progressbar
-    elif verbose == 'tqdm':
-        pbar = tqdm(iterable, desc=desc, leave=leave, position=position,
-                    dynamic_ncols=True)
-    elif verbose == 'tqdm_notebook':
-        from tqdm import tqdm_notebook
-        pbar = tqdm_notebook(iterable, desc=desc, leave=leave,
-                             position=position, dynamic_ncols=True)
-    elif verbose is False:
+def _pbar(iterable, desc, verbose=True, **kwargs):
```
jasmainak, line 168: So the `leave` and `position` options were for nested loops in autoreject. See my [issue on tqdm](https://github.com/tqdm/tqdm/issues/312#issuecomment-261098464). Basically, we want cross-validation fold as the outer loop and `n_interp` nested inside it. It was working before for `tqdm` backend but now it would be broken. Any chance we could get this to work with MNE?

pstetz: These options are covered in `**kwargs` and I'm getting the behavior described in the issue you linked (screenshots below). The `leave` and `position` variables are passed as arguments when this functionality is needed ([`n_interp`](https://github.com/autoreject/autoreject/blob/a60ac1a7a729429d34b075f6644d32604236dc69/autoreject/autoreject.py#L772-L775) and [`Fold`](https://github.com/autoreject/autoreject/blob/a60ac1a7a729429d34b075f6644d32604236dc69/autoreject/autoreject.py#L805-L806))

pstetz: If you like, I can restore the default arguments (`leave=True, position=None`) to the [`_pbar`](https://github.com/autoreject/autoreject/blob/91dee4f7a8d4d76ab3bc2c14c3ff91416dcb513f/autoreject/utils.py#L168)? Sorry if this is so much iteration! It's my first time contributing to open source!

jasmainak: no worries! I think I noticed this now but looks like `tqdm` has changed (or broken?) since the last time I posted that question. It looks like `leave=True` always creates a new line but this was not the case before. It would leave a trace but not create a new line. Try this:

```py
from tqdm import trange
from time import sleep

for i in trange(10, desc='1st loop', leave=True, position=1):
    for j in trange(100, desc='2nd loop', leave=True, position=2):
        sleep(0.01)
    for k in trange(100, desc='3nd loop', leave=True, position=3):
        sleep(0.01)
```

vs 

```py
from tqdm import trange
from time import sleep

for i in trange(10, desc='1st loop', leave=True, position=1):
    for j in trange(100, desc='2nd loop', leave=False, position=2):
        sleep(0.01)
    for k in trange(100, desc='3nd loop', leave=False, position=3):
        sleep(0.01)
```

Neither is the desired behavior that was working before but the second one is closer to what I wanted I suppose (except that the trace should not be erased at the end of the loop)

jasmainak:
@pstetz I guess the `tqdm` issue is unrelated to your PR. I'll have to change `leave=True` to `leave=False` in a couple of places. Do you want to update `whats_new.rst` and then it's good to go for me

PR #248, opened 2021-11-19
[MRG] fix/ci: use machine, not docker in CircleCI

sappelhoff:
follow up to #247 where I made a typo.

Also speeds everything up  :tada:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ jobs:
                     which python3
                     pip3 --version
                     which pip3
-                    mne sys_info
+                    python3 -c "import mne; mne.sys_info()"
 
             - run:
                 name: Download MNE sample data
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,5 @@
-# For an explanation, see: https://github.com/autoreject/autoreject/pull/247#issuecomment-973069510
+# For an explanation of the workflow here, see:
+# https://github.com/autoreject/autoreject/pull/247#issuecomment-973069510
 version: 2.1
 jobs:
     setup:
@@ -8,17 +9,25 @@ jobs:
             - checkout
             - restore_cache:
                 keys:
-                  # restore cache from last build. Unless __init__.py has changed since then
-                  # bump the "vX" from data-cache-vX-* to update cache without changing __init__.py
+                  # restore cache from last build. Unless __init__.py or setup.py have changed since then
+                  # bump the "X" in the "vX" part from data-cache-vX-* to update cache if needed without
+                  # changing __init__.py or setup.py
                   - data-cache-v0-{{ checksum "./autoreject/__init__.py" }}-{{ checksum "setup.py" }}
             - run:
                 name: Setup python3 environment
                 command: |
-                    python3 -m pip install --progress-bar off --upgrade pip
+                    python3 -m venv /home/circleci/autoreject_env
+                    echo "source /home/circleci/autoreject_env/bin/activate" >> $BASH_ENV
+            - run:
+                name: Install packages
+                # setuptools, wheel, psutil not strictly needed
+                command: |
+                    python3 -m pip install --progress-bar off --upgrade pip setuptools wheel psutil
                     python3 -m pip install --progress-bar off -e .[full]
                     python3 -m pip install --progress-bar off -e .[doc]
             - run:
                 name: Print system information
+                # TODO: remove one of the mne calls probably
                 command: |
                     date
                     python3 --version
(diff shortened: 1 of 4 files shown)


jasmainak:
@sappelhoff docbuild1 took 1 hour??

sappelhoff:
yeah, I don't get it. :-(

sappelhoff:
image

it took 14 minutes the last times: https://app.circleci.com/pipelines/github/autoreject/autoreject/218/workflows/0f101d91-2ece-442e-a03b-aa735a926bd8

I'll push one more commit to see whether this was an outlier.

sappelhoff:
it seems like the execution times are a little unstable @jasmainak 

image


I don't know why. could this be a random element in the autoreject examples? Or is this rather due to circleci resources being provided not perfectly stable?

jasmainak:
How do you get that graph?! The variability in CircleCI runs is too much ... 14 minutes to 1 hour. Are you sure that the data downloaded in the "setup" build is being used in subsequent builds? One easy check is to run sphinx gallery in verbose mode and you'll see if it's redownloading the data.

I timed the same example on my 2017 macbook air (8 GB RAM) and I got:

```py
>>> %timeit %run examples/plot_auto_repair.py
2min 57s ± 15.7 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

something feels not right

jasmainak:
Also, we set `random_state` in the examples so there should be absolutely no random element in the executions

sappelhoff:
you can get the graph by clicking on "commit" after clicking on the general circleci check, see screenshot:

image

https://app.circleci.com/pipelines/github/autoreject/autoreject/220/workflows/711684f5-1349-4a9e-ae31-079d2e617051

sappelhoff:
Okay, I am running all of it in verbose mode now and that shows that downloading the data is no problem, see [here](https://app.circleci.com/pipelines/github/autoreject/autoreject/221/workflows/338ee826-d6c2-4b22-97f6-a9c2033316b4/jobs/981?invite=true#step-102-129) 

> Skipping sub-16_ses-meg_task-facerecognition_run-01_meg.fif: **already downloaded**.: 100%|##########| 824M/824M [00:00<?, ?B/s]

It looks more like it sometimes gets a bit "slow" on the `Creating augmented epochs` step :shrug: why could that be slow on CircleCI?

sappelhoff:
And it's always docbuild1 and docbuild7 that take so long

- https://github.com/autoreject/autoreject/blob/master/examples/plot_auto_repair.py
- https://github.com/autoreject/autoreject/blob/master/examples/plot_visualize_bad_epochs.py

jasmainak:
Maybe it's a memory issue. Depending on how much memory CircleCI allocates, it might slow/speed up the job. I'm grabbing at straws a bit but I would go in that direction if I had to dig. Using `memory_profiler` on the code locally on my computer. 

I'm not surprised that `docbuild7` is slow. That example probably needs to go anyway. But would be good to have `docbuild1` work as it is without modifying the example

jasmainak:
Do concurrent builds share the memory? Any idea? Maybe `docbuild7` is affecting `docbuild1`

sappelhoff:
yes, memory could indeed be the issue, looking at the circleci docs: https://circleci.com/docs/2.0/configuration-reference/#resource_class

> We implement soft concurrency limits for each resource class to ensure our system remains stable for all customers.

So sometimes we'll get a bit more RAM that at other times, depending on how heavily the CircleCI servers are used, if I understand that correctly.

Furthermore:

> vailable resources classes are restricted for customers on the Free plan to **small/medium for linux**

image


meaning we get something between a minimum of 2 and a maximum of 4 GB.

Things we could do to confirm that this is indeed the problem:

1. memory profile docbuild1 and docbuild7 examples locally, to check whether they need more than 2GB
2. track memory usage in the circleci jobs: https://support.circleci.com/hc/en-us/articles/360043994872-How-to-record-a-job-s-memory-usage

However I am reasonably confident that that's really the issue, because it would also explain the apparent "randomness" --> since that one failure, the build has successfully [passed two times](https://app.circleci.com/pipelines/github/autoreject/autoreject/221/workflows/338ee826-d6c2-4b22-97f6-a9c2033316b4).

image

sappelhoff:
**Ahaa!** ... or we don't use a "docker executor" (as currently), but a "machine executor" (standard linux machine) instead --> then we seem to automatically get more resources https://circleci.com/docs/2.0/configuration-reference/#machine-executor-linux

image

maybe I'll just try that.

sappelhoff:
yup, I think that was it @jasmainak ... build times are awesome now:

image


I think this system is now quite robust, easily extensible, and of low maintenance burden :-) ready to be merged from my side.

jasmainak:
Thanks a lot @sappelhoff !!! I owe you a few beers at the very least 🍺 🍻

jasmainak:
I feel like the CircleCI troubles started after the switch to CircleCI 2.0 where the concept of "executors" was introduced. Probably all builds prior were machine builds? Anyhow, thanks a lot for debugging. I'm guessing the set up stage can be made even faster through caching :)

jasmainak:
Btw, some sphinx-gallery weirdness, thumbnails are not shown for all examples : with the parallel + caching trick: 

https://1034-59485140-gh.circle-artifacts.com/0/html/auto_examples/index.html

image

sappelhoff:
Mhhh damn. Then maybe we should disable the concurrent/parallel jobs again ... Now that the executor is of type machine, we are guaranteed two cores and 7.5gb ram... So that would probably suffice anyhow.

I am still happy that I learned about all this stuff, so no problem :)

jasmainak:
yes :) and I saw that it takes 30 minutes to download the dataset. If we cache that, we should definitely within 1 hour

sappelhoff:
The dataset is cached, the 30 minutes is migrating the workspace (dataset, project, python venv) to the individual jobs, and yeah - that step would not be necessary anymore if we drop the parallel jobs again

PR #261, opened 2022-02-04
fix import compatability

ktavabi:


agramfort:
@ktavabi Can you avoid reformatting the entire file with black?
besides I am not sure to get the fix. In the 2 try / except blocks you import different symbols

PR #280, opened 2022-07-17
Fix and refactor RANSAC for n_jobs>1

### LINKED ISSUE #277: Possible bug in RANSAC, job distribution calculated wrong

skjerns:
closes #277 
closes #278 

The current implementation of RANSAC has some bugs renderering it faulty&non-functioning for `n_jobs>1`, and I'm not even sure if the case `n_jobs=1` is maybe also affected by this in certain cases.

I introduce and propose the following changes:
* Compute the mappings (interpolations of subsets of channels) with `Parallel`,  `_get_mappings` is actually the slow part of RANSAC and benefits of parallelization the most (at least for MEG data).
* Fix two bugs (list index taken instead of actual channel index, iterating over spaces of n_jobs instead of all epochs) see also #277 and #278 for details, additionally, the corresponding lines have been commented on in the review )
* The only part that needs randomization is `_get_random_subsets`, which does not need to be parallel, so we can just compute it outside of a Parallel context and don't have to worry about random seeds in subprocesses.
* Bad channels should be excluded from the random subset selection processes. This is excluded later anyway by the interpolation function, but it would probably be good to already exclude them during the random sampling process. Else we possibly have a bias if one random subset contains too many bad channels.
* Additionally I added some sanity checks,. warnings and a test with actual data (artificial line noise added)
 
I propose refactoring these changes
* move `_get_channel_type` to utils
* Be more consistent with `verbose`
* move all functions inside the RANSAC class, `Parallel` has no problem with class methods

### DIFF
--- a/autoreject/ransac.py
+++ b/autoreject/ransac.py
@@ -16,44 +16,8 @@ from mne.parallel import check_n_jobs
 from mne.utils import check_random_state
 
 from .utils import _pbar, _handle_picks
-from .utils import _check_data
+from .utils import _check_data, _get_channel_type
 
-
-def _iterate_epochs(ransac, epochs, idxs, ch_subset, verbose):
-    ransac.mappings_ = ransac._get_mappings(epochs, ch_subset)
-    n_channels = len(ransac.picks)
-    corrs = np.zeros((len(idxs), n_channels))
-    for idx, _ in enumerate(_pbar(idxs, desc='Iterating epochs',
-                                  verbose=verbose)):
-        ransac.corr_ = ransac._compute_correlations(
-            epochs.get_data()[idx, ransac.picks])
-        corrs[idx, :] = ransac.corr_
-    return corrs
-
-
-def _get_channel_type(epochs, picks):
-    picked_info = mne.io.pick.pick_info(epochs.info, picks)
-    ch_types_picked = {
-        mne.io.meas_info.channel_type(picked_info, idx)
-        for idx in range(len(picks))}
-    invalid_ch_types_present = [key for key in ch_types_picked
-                                if key not in ['mag', 'grad', 'eeg'] and
-                                key in epochs]
-    if len(invalid_ch_types_present) > 0:
-        raise ValueError('Invalid channel types present in epochs.'
-                         ' Expected ONLY `meg` or ONLY `eeg`. Got %s'
-                         % ', '.join(invalid_ch_types_present))
-
-    has_meg = any(kk in ch_types_picked for kk in ('mag', 'grad'))
-    if 'eeg' in ch_types_picked and has_meg:
-        raise ValueError('Got mixed channel types. Pick either eeg or meg'
-                         ' but not both')
-    if 'eeg' in ch_types_picked:
-        return 'eeg'
-    elif has_meg:
-        return 'meg'
-    else:
-        raise ValueError('Oh no! Your channel type is not known.')
 
 
 class Ransac(object):
@@ -115,28 +79,42 @@ class Ransac(object):
         self.random_state = random_state
         self.verbose = verbose
         self.picks = picks
+        
 
-    def _get_random_subsets(self, info, random_state):
+    def _iterate_epochs(self, epochs, idxs):
+        n_channels = len(self.picks)
+        corrs = np.zeros((len(idxs), n_channels))
+        for i, idx in enumerate(_pbar(idxs, desc='Iterating epochs',
+                                      verbose=self.verbose)):
+            data = epochs.get_data()[idx, self.picks]
+            corrs[i, :] = self._compute_correlations(data)
+        return corrs
+    
+    def _get_random_subsets(self, info):
         """ Get random channels"""
-        # have to set the seed here
-        rng = check_random_state(random_state)
+        # have to set the seed here, as here the only part with randomization
+        # occurs. However, all subsets are precomputed outside of Parallel,
+        # therefore, we can simply compute them once
+        rng = check_random_state(self.random_state)
         picked_info = mne.io.pick.pick_info(info, self.picks)
         n_channels = len(picked_info['ch_names'])
 
         # number of channels to interpolate from
         n_samples = int(np.round(self.min_channels * n_channels))
 
-        # get picks for resamples
-        picks = []
-        for idx in range(self.n_resample):
-            pick = rng.permutation(n_channels)[:n_samples].copy()
-            picks.append(pick)
-
-        # get channel subsets as lists
+        # get picks for resamples, but ignore channels marked as bad
+        bad_chs = info['bads']
+        ch_list = [ch for ch in picked_info['ch_names'] if not ch in bad_chs]
+        assert len(ch_list)>=n_samples, 'too many channels marked as bad,'\
+            'cannot perform interpolation with min_channels={self.min_channels}'
+            
+        # randomly sample subsets of good channels     
         ch_subsets = []
-        for pick in picks:
-            ch_subsets.append([picked_info['ch_names'][p] for p in pick])
-
+        for idx in range(self.n_resample):
+            picks = rng.choice(ch_list, size=n_samples, replace=False)
+            picks = [str(p) for p in picks] # convert from str-array to string
+            ch_subsets.append(picks)
+            
         return ch_subsets
 
     def _get_mappings(self, inst, ch_subsets):
@@ -149,11 +127,12 @@ class Ransac(object):
         pick_to = range(n_channels)
         mappings = []
         # Try different channel subsets
-        for idx in range(len(ch_subsets)):
+        for subset in _pbar(ch_subsets, desc='interpolating channels',
+                            verbose=self.verbose):
             # don't do the following as it will sort the channels!
             # pick_from = pick_channels(ch_names, ch_subsets[idx])
             pick_from = np.array([ch_names.index(name)
-                                  for name in ch_subsets[idx]])
+                                  for name in subset])
             mapping = np.zeros((n_channels, n_channels))
             if self.ch_type == 'meg':
                 mapping[:, pick_from] = _fast_map_meg_channels(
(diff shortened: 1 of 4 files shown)


agramfort:
@skjerns what empirical testing on real data did you do at this point?
Do you get the same result as before with n_jobs? Now n_jobs does not affect result anymore?

skjerns:
> @skjerns what empirical testing on real data did you do at this point? Do you get the same result as before with n_jobs? Now n_jobs does not affect result anymore?

I added another test that checks if results are the same as before with `n_jobs=1`, seems like they are left untouched.

I tested it by inserting strong line noise into the previously used test data. RANSAC now detects these channels as faulty reliably. Additionally there is testing happening that compares different `n_jobs`, and the results are the same for all settings.

autoreject/tests/test_ransac.py
```diff
@@ -50,12 +53,146 @@ def test_ransac():
...
+    for n_jobs in corrs:
+        # simply compare to the last computed value
+        np.testing.assert_allclose(mappings[1], mappings[n_jobs], err_msg=
+                f'n_jobs={n_jobs} did not produce same results as non-parallel')
+        np.testing.assert_allclose(corrs[1], corrs[n_jobs], err_msg=
+                f'n_jobs={n_jobs} did not produce same results as non-parallel')
+        assert bad_chs[1]==bad_chs[n_jobs], \
+                f'n_jobs={n_jobs} did not produce same results as non-parallel'
+
+def test_ransac_backwards_equality():
```
jasmainak, line 154: I think this test is not needed as long as you've checked that the result doesn't change for n_jobs = 1

autoreject/tests/test_ransac.py
```diff
@@ -50,12 +53,146 @@ def test_ransac():
...
+    ransac.mappings_ = mappings
+    epoch_idxs_splits = np.array_split(np.arange(len(epochs)), 2)
+    corr = ransac._iterate_epochs(epochs, np.arange(len(epochs)))
+    corr_sub1 = ransac._iterate_epochs(epochs, epoch_idxs_splits[0])
+    corr_sub2 = ransac._iterate_epochs(epochs, epoch_idxs_splits[1])
+    np.testing.assert_array_equal(corr, np.concatenate([corr_sub1,
+                                                        corr_sub2]))
+
+    # now test across different jobs
+    mappings = {}
```
jasmainak, line 122:
```suggestion
    mappings = dict()
```

just to be a bit more explicit and different from `list()` or `[]`

autoreject/tests/test_ransac.py
```diff
@@ -50,12 +53,146 @@ def test_ransac():
...
+    # amplify some of the channels with noise as sanity check
+    np.random.seed(0)
+    
+    # create some artifically noisy channels with strong line noise
+    mag_idx = mne.pick_types(epochs.info, meg='mag')
+    noisy_idx = np.random.choice(mag_idx, 5, replace=False)
+    noisy_idx = sorted(noisy_idx)
+    
+    times = np.arange(0, epochs.tmax-epochs.tmin + 1/epochs.info['sfreq'], 
+                      1/epochs.info['sfreq'])
```
jasmainak, line 93: do you have a flake8 linter? It seems to not conform to pep8

autoreject/tests/test_ransac.py
```diff
@@ -50,12 +53,146 @@ def test_ransac():
...
     picks = mne.pick_types(raw.info, meg=False, eeg=True, stim=True,
-                           eog=False, exclude=[])
+                            eog=False, exclude=[])
     ransac = Ransac(picks=picks)
     pytest.raises(ValueError, ransac.fit, epochs)
+
+
+        
+def test_ransac_multiprocessing():
+    """test on real data with 5 random channels augmented with strong
```
jasmainak, line 69: docstring should be single line. If you want multiline, first do a single line summary and then multi-line afterwards

autoreject/ransac.py
```diff
@@ -187,28 +165,60 @@ def _compute_correlations(self, data):
         return corr
 
     def fit(self, epochs):
+        """
+        Perform RANSAC on the given epochs.
```
jasmainak, line 169: summary line of docstring should be on L168

jasmainak:
Mostly stylistic nitpicks, I'm happy to merge once that is fixed and you're satisfied with the fix. Please do update `whats_new.rst` as well !

autoreject/tests/test_ransac.py
skjerns: Which one do you mean exactly? the test for same `bad_chs` being found when using different `n_jobs`? Or the backwards equality?

autoreject/tests/test_ransac.py
skjerns: I should do that indeed. No I don't have it active right now

autoreject/tests/test_ransac.py
skjerns: I removed the backwards check again, but now you know it was there, and it was indeed consistent :)

autoreject/tests/test_ransac.py
skjerns: Ah, didn't know, thanks!

skjerns:
Should all be pep8 consistent now - thanks for the review!

autoreject/tests/test_ransac.py
```diff
@@ -59,3 +60,89 @@ def test_ransac():
...
+    corr_sub2 = ransac._iterate_epochs(epochs, epoch_idxs_splits[1])
+    np.testing.assert_array_equal(corr, np.concatenate([corr_sub1,
+                                                        corr_sub2]))
+
+    # now test across different jobs
+    mappings = dict()
+    corrs = dict()
+    bad_chs = dict()
+
+    for n_jobs in [1, 2, -1]:
```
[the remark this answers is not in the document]
skjerns: makes sense! I've changed it to [1,2,3]

autoreject/tests/test_ransac.py
```diff
@@ -59,3 +60,89 @@ def test_ransac():
...
+
+    event_id = {'Visual/Left': 3}
+    tmin, tmax = -0.2, 0.5
+
+    events = mne.find_events(raw)
+    epochs = mne.Epochs(raw, events, event_id, tmin, tmax,
+                        baseline=(None, 0), reject=None, preload=True)
+    # see if results are consistent when using different n_jobs
+    # amplify some of the channels with noise as sanity check
+    np.random.seed(0)
```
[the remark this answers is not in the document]
skjerns: done

agramfort:
@skjerns can you push an empty commit to restart the build of the doc?

skjerns:
The file in question is indeed 404

https://s3.amazonaws.com/openneuro.org/ds000117/sub-16/ses-meg/beh/sub-16_ses-meg_task-facerecognition_events.tsv?versionId=QYeVxEW3jZIuwwbAYj1NVxp6FQS5jkyU

agramfort:
I see the file online
https://openneuro.org/datasets/ds000117/versions/1.0.5/file-display/sub-16:ses-meg:beh:sub-16_ses-meg_task-facerecognition_events.tsv

sappelhoff:
maybe one could raise an issue here: https://github.com/OpenNeuroOrg/openneuro

or it might be a temporary amazon S3 issue and we can try again in a couple of hours.

skjerns:
> maybe one could raise an issue here: https://github.com/OpenNeuroOrg/openneuro
> 
> or it might be a temporary amazon S3 issue and we can try again in a couple of hours.

without the version specifier the file is available on AWS. I'd also suggest to wait, and else to simply use the file without a version specifier and trust that it likely does not change ever?

sappelhoff:
> and I'm not even sure if the case n_jobs=1 is maybe also affected by this in certain cases.

do you have an update on this? was `n_jobs=1` also affected?

> Fix several bugs (list index taken instead of actual channel index, iterating over n_jobs instead of the entire epochs, etc

I would appreciate a concise list of bugs fixed (with line refs / permalinks to the code prior to this PR) --> currently it's all a bit hard to spot amidst the refactoring

>  I'd also suggest to wait, and else to simply use the file without a version specifier and trust that it likely does not change ever?

+1 to wait ... but -1 to drop the version specifier: let's rather open an openneuro issue first if that keeps being problematic.

autoreject/ransac.py
```diff
@@ -187,28 +166,61 @@ def _compute_correlations(self, data):
...
+            ``self.bad_chs_``
+        """
         self.picks = _handle_picks(info=epochs.info, picks=self.picks)
         _check_data(epochs, picks=self.picks,
                     ch_constraint='single_channel_type', verbose=self.verbose)
         self.ch_type = _get_channel_type(epochs, self.picks)
         n_epochs = len(epochs)
 
         n_jobs = check_n_jobs(self.n_jobs)
-        parallel = Parallel(n_jobs, verbose=10)
```
skjerns, line 197: verbosity parameter should be regarded

autoreject/ransac.py
```diff
@@ -187,28 +166,61 @@ def _compute_correlations(self, data):
...
+        self.mappings_ = np.concatenate(mappings)
+
+        # compute correlations with possibility of parallelization
+        delayed_func = delayed(self._iterate_epochs)
+        n_splits = min(n_jobs, n_epochs)  # max n_epochs splits possible
+        epoch_idxs_splits = np.array_split(np.arange(n_epochs), n_splits)
+        corrs = parallel(delayed_func(epochs, idxs) for idxs
+                         in epoch_idxs_splits)
         self.corr_ = np.concatenate(corrs)
-        if self.verbose is not False and self.n_jobs > 1:
```
skjerns, line 211: Changed this to regard verbosity setting

skjerns:
> do you have an update on this? was `n_jobs=1` also affected?

I don't think it is.


> I would appreciate a concise / exact list of bugs fixed (with line refs / permalinks to the code prior to this PR) --> currently it's all a bit hard to spot amidst the refactoring. The "etc." in your current list of bugs makes me anxious :-)

You are right, I added the specific parts as a comment in the code.

There were three bugs:
* Getting epoch data via enumeration index instead of epoch index
* Splitting up epochs into lists with spacing `n_jobs` instead of number of splits
* Bad channels were incorperated into subchannel sampling (even though they are ignored in the interpolation later, so only a minor bug)

> > I'd also suggest to wait, and else to simply use the file without a version specifier and trust that it likely does not change ever?
> 

 seems reasonable

jasmainak:
we have another PR affected by the same fetching issue ... otherwise I'm happy to merge

jasmainak:
Thanks @skjerns for the fix! 🎉

PR #326, opened 2023-11-23
drop support for py<3.9

sappelhoff:
mne does not support python <3.9, so I bumped dependencies here as well.

I also bumped the minimum mne version to 1.5 ... following the same scheme as in mne-bids --> supporting the latest 2 mne releases (1.5 and 1.6 currently)


setup.cfg
```diff
@@ -20,3 +20,6 @@ addopts =
     --durations=10 -ra --cov-report= --tb=short
     --doctest-ignore-import-errors --junit-xml=junit-results.xml
     --color=yes --capture=sys
+filterwarnings =
+    error
+    ignore:datetime\.datetime\.utcfromtimestamp.* is deprecated and scheduled for removal in a future version.*:DeprecationWarning
```
sappelhoff, line 25: this is not working and I don't get it ... can't seem to find the problem locally either.

Passing `-W ignore::DeprecationWarning` on the command line works 🤷 

any idea @larsoner?

larsoner: I don't think the dot includes newlines so if the warning has them they have to be added

larsoner:
```suggestion
filterwarnings =
    error
    ignore:datetime\.datetime\.utcfromtimestamp.* is deprecated and scheduled for removal in a future version.*\n.*:DeprecationWarning
```

setup.cfg
```diff
@@ -20,3 +20,6 @@ addopts =
     --durations=10 -ra --cov-report= --tb=short
     --doctest-ignore-import-errors --junit-xml=junit-results.xml
     --color=yes --capture=sys
+filterwarnings =
+    error
+    ignore:datetime\.datetime\.utcfromtimestamp.* is deprecated and scheduled for removal in a future version.*\n.*:DeprecationWarning
```
larsoner, line 25:
```suggestion
    ignore:(\n|.)*datetime\.datetime\.utcfromtimestamp.* is deprecated and(\n|.)*:DeprecationWarning
```

.github/workflows/test.yml
```diff
@@ -121,7 +121,7 @@ jobs:
     - name: Clone MNE-Python
       run: git clone --depth 1 --single-branch --branch main https://github.com/mne-tools/mne-python.git
```
larsoner, line 122: Depth 1 breaks version number
```suggestion
      run: git clone --single-branch --branch main https://github.com/mne-tools/mne-python.git
```

larsoner:
The `(\n|.)` trick worked in MNE-Python but doesn't appear to work here. Tomorrow I can look locally (just tried pushing without testing locally so far) and figure it out. Might be worth using pytest_configure in conftest like we do in MNE-Python to rule out parsing issues with setup.cfg and escapes for example

autoreject/conftest.py
```diff
@@ -28,6 +28,7 @@ def pytest_configure(config):
     ignore:FigureCanvasAgg is non-interactive.*:UserWarning
     ignore:`np.MachAr` is deprecated.*:DeprecationWarning
     ignore:`product` is deprecated as of NumPy.*:DeprecationWarning
+    ignore:datetime\.datetime\.utcfromtimestamp.* is deprecated and.*:DeprecationWarning
```
larsoner, line 31: @sappelhoff it turned out we were already using `pytest_configure` to set the ignore lines, so setting them in `setup.cfg` (as well) did not have the desired effect.

sappelhoff:
Ok, thanks a lot for looking into it, Eric!

I'll be switching mne-bids over to pyproject.toml soon and will then do the same here to move away from setup.cfg, which seems less and less well supported by tools.

autoreject/conftest.py
sappelhoff: ohhh thanks @larsoner, I should have checked that!

PR #347, opened 2025-12-16
feat: add GPU acceleration via PyTorch backend

Ramdam17:
Add optional GPU acceleration for AutoReject using PyTorch, providing 10x+ speedups on large EEG/MEG datasets while maintaining full backward compatibility.

New modules:
- backends.py: Unified interface for NumPy/PyTorch compute backends
- gpu_interpolation.py: GPU-optimized spherical spline interpolation
- gpu_pipeline.py: End-to-end GPU pipeline with data-on-device architecture

Key features:
- Automatic hardware detection (CUDA, MPS, CPU)
- Opt-in activation via AUTOREJECT_BACKEND=torch environment variable
- Numerical consistency with CPU implementation (<0.1% difference)
- Graceful fallback when PyTorch is not installed

Installation: pip install autoreject[gpu]


autoreject/autoreject.py
```diff
@@ -583,13 +626,41 @@ def _vote_bad_epochs(self, epochs, picks):
...
+        labels : ndarray
+            Label array (0=good, 1=bad).
+        picks : array-like
+            Channel indices to process.
+        n_interpolate : int
+            Maximum number of channels to interpolate.
+        verbose : bool
+            Verbosity.
+        data : ndarray or None
+            Pre-extracted data array (n_epochs, n_channels, n_times).
```
jasmainak: is this a single epoch or multiple epochs?

autoreject/autoreject.py
```diff
@@ -583,13 +626,41 @@ def _vote_bad_epochs(self, epochs, picks):
...
+        
+        # Pre-compute data for all epochs if we'll need it for PTP computation
+        # This avoids repeated epochs[idx].get_data() calls
+        data_local = data
+        if data_local is None:
+            # Check if any epoch needs PTP computation (n_bads > n_interpolate)
+            n_bads_per_epoch = np.sum(labels[:, picks] == 1, axis=1)
+            needs_ptp = np.any(n_bads_per_epoch > n_interpolate)
+            if needs_ptp:
+                data_local = epochs.get_data(**_GDKW)
```
jasmainak: where is _GDKW defined?

autoreject/autoreject.py
```diff
@@ -599,8 +670,13 @@ def _get_epochs_interpolation(self, epochs, labels,
                     interp_chs_mask = labels[epoch_idx] == 1
                 else:
                     # get peak-to-peak for channels in that epoch
-                    data = epochs[epoch_idx].get_data(**_GDKW)[0]
-                    peaks = np.ptp(data, axis=-1)
+                    if data_local is not None:
+                        epoch_data = data_local[epoch_idx]
+                    else:
+                        # Fallback: should not happen if needs_ptp logic is correct
```
jasmainak: raise error if it should not happen? Your fix may silently suppress another issue ...

autoreject/autoreject.py
```diff
@@ -767,24 +846,203 @@ def transform(self, epochs, return_log=False):
...
+        Precomputed dots for interpolation.
+    picks : array-like
+        Channel indices to consider.
+    
+    Returns
+    -------
+    interpolated_data : ndarray, shape (n_channels, n_times)
+        The interpolated epoch data.
+    """
+    from mne import EpochsArray
```
jasmainak: import globally above?

autoreject/autoreject.py
```diff
@@ -767,24 +846,203 @@ def transform(self, epochs, return_log=False):
...
+        For each epoch, list of channel names to interpolate.
+    picks : array-like
+        Channel indices to consider.
+    device : str or None
+        GPU device to use.
+    verbose : bool
+        Whether to show progress bar.
+    """
+    from .gpu_interpolation import gpu_interpolate_bad_epochs
+    from .backends import get_backend
```
jasmainak, line 906: import at top of file?

autoreject/autoreject.py
```diff
@@ -767,24 +846,203 @@ def transform(self, epochs, return_log=False):
...
+        device = backend.device
+    
+    picks = np.asarray(picks)
+    n_picks = len(picks)
+    
+    # Get positions for picked channels
+    pos = epochs._get_channel_positions(picks)
+    # Normalize positions
+    norms = np.linalg.norm(pos, axis=1, keepdims=True)
+    pos = pos / norms
```
jasmainak:
```suggestion
    pos /= norms
```

autoreject/autoreject.py
```diff
@@ -767,24 +846,203 @@ def transform(self, epochs, return_log=False):
...
+                list(enumerate(interp_channels)),
+                desc='Repairing epochs (parallel)',
+                position=pos, leave=True, verbose=verbose
+            )
+        )
+        
+        # Update epochs data with interpolated results
+        for epoch_idx, interpolated_data in enumerate(results):
+            epochs._data[epoch_idx] = interpolated_data
+    else:
```
jasmainak: do you need the if/else? Doesn't joblib fallback to the original behavior if n_jobs = 1?

autoreject/autoreject.py
```diff
@@ -767,24 +846,203 @@ def transform(self, epochs, return_log=False):
...
+    cv : sklearn cross-validator
+        Cross-validation object.
+    consensus : array-like
+        Values of consensus to try.
+    dots : tuple | None
+        Precomputed dots for interpolation.
+    verbose : bool
+        Whether to show progress.
+    n_jobs : int
+        Number of parallel jobs for interpolation. Default 1.
```
jasmainak, line 1021: is the behavior of n_jobs same as before? Was it parallel across cv folds before?

autoreject/autoreject.py
```diff
@@ -918,12 +1177,16 @@ class AutoReject:
     picks_ : array-like, shape (n_data_channels,)
         The data channels considered by autoreject. By default
         only data channels, not already marked as bads are considered.
+    device : str
+        Compute device: 'auto' (default), 'gpu', 'mps', 'cuda', or 'cpu'.
```
jasmainak: for my curiosity what is the difference between cuda and gpu? is there a way for the user to know which one to select if 'auto' does not work?

autoreject/autoreject.py
```diff
@@ -933,10 +1196,64 @@ def __init__(self, n_interpolate=None, consensus=None,
...
+        ----------
+        epochs : mne.Epochs
+            The epochs to process.
+        
+        Returns
+        -------
+        use_gpu : bool
+            Whether to use GPU.
+        device : str
+            Device to use ('mps', 'cuda', or 'cpu').
```
jasmainak: `'gpu'` is not an option here?

autoreject/backends.py
```diff
@@ -0,0 +1,931 @@
...
+import numpy as np
+
+
+__all__ = ['detect_hardware', 'get_backend', 'get_backend_names', 
+           'DeviceArray', 'is_device_array', 'force_cpu_backend']
+
+
+# =============================================================================
+# DeviceArray - Wrapper for GPU-resident arrays
+# =============================================================================
```
jasmainak: could you align the == with the text length

autoreject/gpu_pipeline.py
```diff
@@ -0,0 +1,1364 @@
+"""
```
jasmainak, line 1: Add your name and email at the top of the file!

doc/whats_new.rst
```diff
@@ -15,6 +15,8 @@ What's new?
 Changelog
 ~~~~~~~~~
 
+- Add optional GPU acceleration via PyTorch backend (CUDA and Apple MPS). 
+  Set ``AUTOREJECT_BACKEND=torch`` environment variable to enable.
```
jasmainak: add your name and pull request number too!

README.rst
```diff
@@ -69,6 +69,20 @@ All required dependencies are installed automatically when you install ``autorej
...
+GPU Acceleration
+----------------
+
+For faster processing on large datasets, GPU acceleration is available via PyTorch::
+
+    pip install autoreject[gpu]
+
+To enable GPU acceleration, set the backend via environment variable::
+
+    export AUTOREJECT_BACKEND=torch
```
jasmainak, line 83: would it be possible to set it within python with a context manager? Like MNE does for `use_3d_backend` or `set_3d_backend`

autoreject/tests/test_gpu_interpolation.py
```diff
@@ -0,0 +1,253 @@
...
+    clear_backend_cache()
+    
+    yield
+    
+    # Restore original value (not just delete!)
+    if original_backend is None:
+        os.environ.pop('AUTOREJECT_BACKEND', None)
+    else:
+        os.environ['AUTOREJECT_BACKEND'] = original_backend
+    clear_backend_cache()
```
jasmainak: definitely a context manager would help!

autoreject/tests/test_gpu_interpolation.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+from autoreject.gpu_interpolation import (
+    legval_torch,
+    _calc_g_torch,
+    _normalize_vectors_torch,
+    gpu_make_interpolation_matrix,
+    gpu_do_interp_dots,
+)
```
jasmainak: move imports to top of file?

autoreject/tests/test_gpu_interpolation.py
```diff
@@ -0,0 +1,253 @@
...
+        """Test that output has correct shape."""
+        pos_good = np.random.randn(20, 3)
+        pos_bad = np.random.randn(3, 3)
+        
+        result = gpu_make_interpolation_matrix(pos_good, pos_bad)
+        
+        assert result.data.shape == (3, 20)
+
+
+class TestGpuDoInterpDots:
```
jasmainak:
```suggestion
class TestGPUDoInterpDots:
```

autoreject/tests/test_backends.py
```diff
@@ -0,0 +1,641 @@
...
+        assert isinstance(zeros, DeviceArray)
+        
+        result_sum = backend.sum(gpu_data, axis=0, keep_on_device=True)
+        assert isinstance(result_sum, DeviceArray)
+        
+        result_mean = backend.mean(gpu_data, axis=0, keep_on_device=True)
+        assert isinstance(result_mean, DeviceArray)
+        
+        result_sqrt = backend.sqrt(backend.abs(gpu_data, keep_on_device=True), keep_on_device=True)
+        assert isinstance(result_sqrt, DeviceArray)
```
jasmainak, line 611: This file looks pretty thorough! Do you need a test that checks the same output on the full autoreject (local) pipeline with different backends? It can be with a small number of epochs and channels ...

autoreject/tests/test_backends.py
```diff
@@ -0,0 +1,641 @@
...
+        """Test that CPU is always detected as available."""
+        from autoreject.backends import detect_hardware, clear_backend_cache
+        
+        clear_backend_cache()
+        hw = detect_hardware()
+        assert hw.get('cpu') is True
+    
+    def test_detect_hardware_cached(self):
+        """Test that detect_hardware results are cached."""
+        from autoreject.backends import detect_hardware, clear_backend_cache
```
jasmainak: can be a import at top of file?

autoreject/gpu_pipeline.py
```diff
@@ -0,0 +1,1364 @@
...
+        # Even: return average of two middle elements
+        idx1, idx2 = n // 2 - 1, n // 2
+        v1 = torch.select(sorted_t, dim, idx1)
+        v2 = torch.select(sorted_t, dim, idx2)
+        return (v1 + v2) / 2
+
+
+class GPUThresholdOptimizer:
+    """
+    GPU-accelerated threshold optimizer for autoreject.
```
jasmainak:
```suggestion
    """GPU-accelerated threshold optimizer for autoreject.
```

autoreject/gpu_pipeline.py
```diff
@@ -0,0 +1,1364 @@
...
+        -------
+        ptp : torch.Tensor, shape (n_epochs, n_channels)
+            Peak-to-peak values on GPU.
+        """
+        data_gpu = self._to_tensor(data)
+        return data_gpu.max(dim=-1).values - data_gpu.min(dim=-1).values
+    
+    def batched_channel_cv_loss(self, data_1d, thresholds, cv_splits, y=None):
+        """
+        Compute cross-validated loss for ALL thresholds at once for a single channel.
```
jasmainak:
```suggestion
        """Compute cross-validated loss for ALL thresholds at once for a single channel.
```

autoreject/gpu_pipeline.py
```diff
@@ -0,0 +1,1364 @@
...
+            
+            fold_losses[fold_idx] = rmse
+        
+        # Mean across folds (this is what cross_val_score returns, negated)
+        return fold_losses.mean(dim=0)
+    
+    def batched_all_channels_cv_loss_parallel(self, data_all_channels, ptp_all, 
+                                               threshes_all, cv_splits):
+        """
+        Compute cross-validated loss for ALL channels and ALL thresholds at once.
```
jasmainak:
```suggestion
        """Compute cross-validated loss for ALL channels and ALL thresholds at once.
```

autoreject/gpu_pipeline.py
```diff
@@ -0,0 +1,1364 @@
...
+        
+        FULLY PARALLEL VERSION: uses BMM instead of 4D broadcast for memory efficiency
```
jasmainak:
```suggestion
        FULLY PARALLEL VERSION: uses torch.bmm() instead of 4D broadcast for memory efficiency
```

autoreject/gpu_pipeline.py
```diff
@@ -0,0 +1,1364 @@
...
+    
+    optimizer.clear_cache()
+    
+    if verbose:
+        print("  ✓ Batch processing complete!")
+    
+    return local_reject, loss
+
+
+# Utility function to check if GPU should be used
```
jasmainak:
```suggestion
```

use docstring ...

autoreject/__init__.py
```diff
@@ -12,3 +12,18 @@
 from .autoreject import compute_thresholds, validation_curve, get_rejection_threshold
 from .ransac import Ransac
 from .utils import set_matplotlib_defaults
+from .backends import (
+    detect_hardware, get_backend, get_backend_names,
+    DeviceArray, is_device_array
+)
```
jasmainak, line 18: anything to expose in `api_reference.py` for the user?

jasmainak (commented):
@Ramdam17 I want to congratulate you on this substantial effort! 🎉 It is clearly a significant improvement for autoreject users, so thanks for pushing these changes upstream. You have my blessing, wondering if @agramfort and @dengemann have any opinions. 

 It looks like you need a rebase for the code to be merge-ready. I will attempt to review the code in the coming days ... most comments will likely be stylistic and I will run it locally on my computer once.If that works, then you will have my +1. As we discussed offline, I'm counting on you to be the point person for GPU-related bug reports.

dengemann:
Wow. It seems we arrived in the future 😀🎉 - amazing 🤩

autoreject/autoreject.py
Ramdam17: It's for **multiple epochs**. I've clarified the docstring to make this explicit:

```python
data : ndarray or None
    Pre-extracted data array for MULTIPLE epochs, shape (n_epochs, n_channels, n_times).
    If provided, avoids expensive repeated ``epochs[idx].get_data()`` calls.
    If None, data will be extracted as needed.
This parameter allows passing all epochs data at once to avoid the overhead of calling epochs[idx].get_data() repeatedly inside the loop.

File: autoreject/autoreject.py (hunk 3 above)
Comment (Ramdam17) [reply autoreject/autoreject.py]: Good catch, thank you! Changed to raise an explicit error:

```python
else:
    # This should not happen if needs_ptp logic is correct
    raise RuntimeError(
        "Unexpected fallback to individual epoch data extraction. "
        "This indicates a bug in needs_ptp computation logic."
    )
```

autoreject/autoreject.py
Ramdam17: Done! Moved `EpochsArray` to the top with other MNE imports since MNE is a required dependency:

```python
import mne
from mne import pick_types, EpochsArray
from mne.viz import plot_epochs as plot_mne_epochs
```

Note: I kept torch-related imports lazy (inside functions) since PyTorch is a 500MB optional dependency. Moving them to module level would cause ImportError for users without torch installed, even when using CPU-only mode.

autoreject/autoreject.py
Ramdam17: These imports need to stay lazy (inside the function) because they depend on PyTorch, which is a 500MB optional dependency.

Moving them to module level would cause `ImportError` for users without PyTorch installed, even when they only want to use CPU-only mode. The lazy import pattern ensures the package works for everyone and only requires PyTorch when GPU acceleration is explicitly requested.

Same pattern applies to other torch-related imports throughout the file (`gpu_interpolation`, `gpu_pipeline`, etc.).

autoreject/autoreject.py
Ramdam17: You're right! Joblib's overhead for `n_jobs=1` is minimal. I've simplified the code to always use `Parallel`:

```python
pos = 2

# Parallel execution (joblib handles n_jobs=1 efficiently)
parallel = Parallel(n_jobs=n_jobs, verbose=0)
my_interp = delayed(_interpolate_single_epoch)

results = parallel(
    my_interp(
        epochs._data[epoch_idx].copy(),
        epochs.info,
        interp_chs,
        dots,
        picks
    )
    for epoch_idx, interp_chs in _pbar(
        list(enumerate(interp_channels)),
        desc='Repairing epochs',
        position=pos, leave=True, verbose=verbose
    )
)

# Update epochs data with interpolated results
for epoch_idx, interpolated_data in enumerate(results):
    epochs._data[epoch_idx] = interpolated_data
```  
  
I initially added this optimization to avoid the joblib overhead when I was looking into parallelization with Numba, but removing the branch simplifies the code significantly as you suggested.

autoreject/autoreject.py
Ramdam17: The behavior has slightly changed:

**Before**: `n_jobs` was only used in `_compute_thresholds()` to parallelize threshold computation across channels.

**Now**: `n_jobs` is used in two places:
1. `_compute_thresholds()` - parallelize across channels (same as before)
2. `_interpolate_bad_epochs()` - **additionally** parallelize epoch interpolation within each CV fold

CV folds themselves are still processed sequentially (not parallelized), but the epoch interpolation within each fold now benefits from parallel processing when `n_jobs > 1`. This provides additional speedup without changing the CV logic.

autoreject/autoreject.py
Ramdam17: Here's the difference:

- **`'cuda'`**: Explicitly requests NVIDIA GPU via CUDA. Will fail if no CUDA-capable GPU is available.
- **`'gpu'`**: Auto-detects the best available GPU (tries CUDA first, then falls back to MPS on Apple Silicon). More forgiving.
- **`'auto'`**: Goes further by considering dataset size heuristics (only uses GPU if >50 epochs, since small datasets don't benefit from GPU due to ~200ms transfer overhead).

I've expanded the docstring to clarify this:

```python
device : str
    Compute device for GPU acceleration. Options:

    - 'auto' (default): Automatically selects GPU if available and beneficial
      for the dataset size (>50 epochs)
    - 'cpu': Force CPU-only computation (uses NumPy backend)
    - 'cuda': Force NVIDIA GPU (requires CUDA-capable GPU and PyTorch with CUDA)
    - 'mps': Force Apple Silicon GPU (requires Apple Silicon Mac and PyTorch with MPS)
    - 'gpu': Auto-detect best available GPU (tries CUDA first, then MPS)

    If PyTorch is not installed, always falls back to CPU. GPU acceleration
    provides 10x+ speedups on large datasets but incurs ~200ms transfer overhead,
    so 'auto' only uses GPU when beneficial.
```

Users can check their available hardware with:

```python
from autoreject.backends import detect_hardware
print(detect_hardware())  # {'cpu': True, 'cuda': False, 'mps': True, ...}
```

autoreject/autoreject.py
Ramdam17: Fixed the docstring to include all options:

```python
device : str
    Device to use ('auto', 'gpu', 'mps', 'cuda', or 'cpu').
```    
    
You're right that 'gpu' and 'auto' can be passed here, even though they get resolved to specific devices ('cuda' or 'mps') later in the logic.

autoreject/backends.py
Ramdam17: Done! Aligned all section separators with their text length throughout the file.

autoreject/gpu_pipeline.py
Ramdam17: Done.

doc/whats_new.rst
Ramdam17: Done.

README.rst
Ramdam17: Great suggestion! I've added a `use_backend()` context manager following MNE's pattern:

```python
from autoreject import AutoReject
from autoreject.backends import use_backend

with use_backend('torch'):
    ar = AutoReject()
    epochs_clean = ar.fit_transform(epochs)
# Automatically reverts to default backend after the context
```

I updated the README to show both options (environment variable and context manager), with the context manager as the recommended approach. The implementation properly saves and restores the backend state, just like MNE's use_3d_backend().

autoreject/tests/test_gpu_interpolation.py
Ramdam17: Exactly! The new `use_backend()` context manager simplifies this significantly. I've updated the test fixtures:

**Before** (manual env management):
```python
@pytest.fixture(autouse=True)
def setup_torch_backend():
    original_backend = os.environ.get('AUTOREJECT_BACKEND')
    os.environ['AUTOREJECT_BACKEND'] = 'torch'
    clear_backend_cache()
    yield
    if original_backend is None:
        os.environ.pop('AUTOREJECT_BACKEND', None)
    else:
        os.environ['AUTOREJECT_BACKEND'] = original_backend
    clear_backend_cache()
```
    
After (context manager):

```python
@pytest.fixture(autouse=True)
def setup_torch_backend():
    """Force torch backend for all tests in this module."""
    with use_backend('torch'):
        yield
```


I updated both test_gpu_interpolation.py and test_backends.py.

autoreject/tests/test_gpu_interpolation.py
Ramdam17: These imports are intentionally placed after `pytest.importorskip("torch")` at line 7. If we move them to the top, the test file would fail to import on systems without PyTorch installed, preventing pytest from even collecting the tests.

The current pattern is:
1. Line 7: `torch = pytest.importorskip("torch")` - Skip module if torch not available
2. Lines 10-16: Import stable dependencies (`use_backend`, `is_device_array`)
3. Lines 20-26: Import torch-dependent modules (`gpu_interpolation` functions)

This ensures clean test skipping on systems without PyTorch. This is the standard pytest pattern for optional dependencies.

A ruff warning is expected for this pattern and can be safely ignored in test files with optional dependencies.

autoreject/tests/test_backends.py
Ramdam17: I've added an test that verifies AutoReject produces consistent results between numpy and torch backends:

```python
def test_autoreject_consistency_across_backends(self):
    """Test AutoReject produces consistent results with numpy vs torch backends."""
    # Create synthetic EEG data (20 epochs, 8 channels)
    
    # Run with numpy backend
    with use_backend('numpy'):
        ar_numpy = AutoReject(n_interpolate=[1, 2], consensus=[...], cv=3, random_state=42)
        ar_numpy.fit(epochs)
        threshes_numpy = ar_numpy.threshes_
        
    # Run with torch backend  
    with use_backend('torch'):
        ar_torch = AutoReject(n_interpolate=[1, 2], consensus=[...], cv=3, random_state=42)
        ar_torch.fit(epochs)
        threshes_torch = ar_torch.threshes_
        
    # Compare: thresholds should be very close (within 0.1% tolerance)
    for ch_name in threshes_numpy.keys():
        assert_allclose(threshes_numpy[ch_name], threshes_torch[ch_name], rtol=0.001)
    
    # Consensus and n_interpolate should be identical (discrete values)
    assert consensus_numpy == consensus_torch
    assert n_interp_numpy == n_interp_torch
```

The test confirms that both backends produce nearly identical results. With the test data, thresholds match to machine precision (0.000% difference), and consensus/n_interpolate are bitwise identical. The 0.1% tolerance accounts for potential float32 (MPS) vs float64 (CPU/CUDA) precision differences on different hardware.

autoreject/tests/test_backends.py
Ramdam17: Done! Moved all stable imports to the top of the file:

```python
from autoreject.backends import (
    detect_hardware,
    clear_backend_cache,
    get_backend,
    get_backend_names,
    NumpyBackend,
    TorchBackend,
    DeviceArray,
    use_backend,
)
```

Removed all redundant inline imports from individual test methods. The only remaining inline imports are in the end-to-end test (test_autoreject_consistency_across_backends) for mne and AutoReject, which are only used in that specific test.

autoreject/__init__.py
Ramdam17: I added a "Compute Backends" section to `doc/api.rst` in commit 11bd8fdf that exposes the user-facing functions:
- `use_backend()` - context manager for backend switching
- `get_backend()` - get current backend
- `detect_hardware()` - check available hardware
- `get_backend_names()` - list available backends

`DeviceArray` and `is_device_array` are lower-level utilities that users typically won't need directly, so I kept them out of the main API docs.

autoreject/autoreject.py
```diff
@@ -565,42 +678,105 @@ def _vote_bad_epochs(self, epochs, picks):
...
                     # get peak-to-peak for channels in that epoch
-                    data = epochs[epoch_idx].get_data(**_GDKW)[0]
-                    peaks = np.ptp(data, axis=-1)
+                    if data_local is not None:
+                        epoch_data = data_local[epoch_idx]
+                    else:
+                        # This should not happen if needs_ptp logic is correct
+                        raise RuntimeError(
+                            "Unexpected fallback to individual epoch data extraction. "
+                            "This indicates a bug in needs_ptp computation logic."
```
jasmainak: seems like CircleCI didn't pass here: https://output.circle-artifacts.com/output/job/c33d2fc6-89fa-4cdf-9c13-0d5f66369674/artifacts/0/html/auto_examples/plot_autoreject_workflow.html

autoreject/autoreject.py
```diff
@@ -746,76 +937,268 @@ def transform(self, epochs, return_log=False):
...
+
+    backend = get_backend()
+    if backend.name != "torch":
+        raise RuntimeError("_interpolate_bad_epochs_gpu requires torch backend")
+
+    if device is None:
+        device = backend.device
+
+    picks = np.asarray(picks)
+    len(picks)
```
jasmainak:
```suggestion
```

autoreject/autoreject.py
```diff
@@ -746,76 +937,268 @@ def transform(self, epochs, return_log=False):
...
+    # Try GPU path if requested
+    if use_gpu:
+        try:
+            _interpolate_bad_epochs_gpu(
+                epochs, interp_channels, picks, device=device, verbose=verbose
+            )
+            return
+        except Exception as e:
+            if verbose:
+                print(f"GPU interpolation failed, falling back to CPU: {e}")
```
jasmainak, line 978: just wondering why not check if GPU available at the beginning and either raise an error or continue? If it does fail, I am guessing it will fail for all epochs resulting in an overly verbose output ...

autoreject/tests/test_backends.py
```diff
@@ -0,0 +1,711 @@
+"""Tests for compute backends.
+
+These tests verify that all backends produce correct and consistent results.
+They also test the fallback mechanisms and environment variable configuration.
+"""
+
+# Authors: autoreject contributors
```
jasmainak: do you mind adding your name and email instead? I know MNE has moved to just writing a generic name but since we only have a few contributors in autoreject, we can keep individual names so people know who to contact.

autoreject/tests/test_backends.py
```diff
@@ -0,0 +1,711 @@
...
+                err_msg=f"Thresholds differ for channel {ch_name}"
+            )
+
+        # Consensus and n_interpolate should be identical
+        # (same CV logic, discrete values)
+        assert consensus_numpy == consensus_torch, (
+            f"Consensus differs: {consensus_numpy} vs {consensus_torch}"
+        )
+        assert n_interp_numpy == n_interp_torch, \
+            f"n_interpolate differs: {n_interp_numpy} vs {n_interp_torch}"
```
jasmainak, line 711: is there a place where you also test the rest of the pipeline ... the reject_log and reject_log.plot() etc? Just to ensure that you have everything necessary in the object that is outputted by the GPU pipeline.

autoreject/tests/test_gpu_interpolation.py
jasmainak: could `setup_torch_backend` be after the imports in line 19-26 or does the said pattern require that fixture to be in between the two imports?

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+max-line-length = 88
+exclude = __init__.py,.git,__pycache__,build,dist,.eggs,*.egg-info,.venv,venv
+ignore = E241,W504
```
jasmainak: can this be added to `pyproject.toml` ?

.claude/settings.local.json
```diff
@@ -0,0 +1,7 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(.venv/bin/python:*)"
+    ]
+  }
+}
```
jasmainak: I think it is a user-specific setting that should not be committed to git?

autoreject/autoreject.py
```diff
@@ -44,8 +60,15 @@ def _slicemean(obj, this_slice, axis):
...
-                     cv=None, return_param_range=False, n_jobs=1):
+def validation_curve(
+    epochs,
+    y=None,
+    param_name="thresh",
+    param_range=None,
+    cv=None,
+    return_param_range=False,
+    n_jobs=1,
+):
```
jasmainak: I know some automatic formatting tools do this but I actually find the original version more readable. Is it possible to easily revert to that version? It will also make it slightly easier for me to see what are the actual code changes independent of formatting changes.

autoreject/autoreject.py
```diff
@@ -746,76 +937,268 @@ def transform(self, epochs, return_log=False):
...
     local_reject.fit(epochs)
 
     assert len(local_reject.consensus_) == 1  # works with one ch_type
     ch_type = next(iter(local_reject.consensus_))
 
-    labels, bad_sensor_counts = \
-        local_reject._vote_bad_epochs(epochs, picks=picks_)
-    desc = 'n_interp'
+    labels, bad_sensor_counts = local_reject._vote_bad_epochs(epochs, picks=picks_)
+    desc = "n_interp"
```
jasmainak: request you to revert to original formatting where possible so it's easier to evaluate the code logic.

autoreject/autoreject.py
```diff
@@ -918,12 +1303,37 @@ class AutoReject:
...
+        - 'cuda': Force NVIDIA GPU (requires CUDA-capable GPU and
+          PyTorch with CUDA)
+        - 'mps': Force Apple Silicon GPU (requires Apple Silicon Mac
+          and PyTorch with MPS)
+        - 'gpu': Auto-detect best available GPU (tries CUDA first, then MPS)
+
+        If PyTorch is not installed, always falls back to CPU.
+        GPU acceleration provides 10x+ speedups on large datasets but
+        incurs ~200ms transfer overhead, so 'auto' only uses GPU when
+        beneficial.
```
jasmainak, line 1179: This is great!!

jasmainak (commented):
Thanks for addressing the comments so far @Ramdam17 ! We are indeed entering the future as @dengemann said.

autoreject/autoreject.py
Ramdam17: Thanks for catching this one! The check `n_bads_per_epoch = np.sum(labels[:, picks] == 1, axis=1)` was only looking at picked channels, but `interp_chs_mask` operates on all channels. This caused false negatives where some epochs needed PTP computation but `needs_ptp=False`, leaving `data_local=None` and triggering the fallback RuntimeError.

Fix: Changed to `n_bads_per_epoch = np.sum(labels == 1, axis=1)` to check all channels in latest commit d730089).

The RuntimeError did exactly what it was supposed to do: catch a logic bug before it caused silent data corruption (and thanx again for this one also :D).

I've checked that `plot_autoreject_workflow.py` now runs successfully end-to-end.

autoreject/autoreject.py
Ramdam17: Actually, the try/except wraps the entire `_interpolate_bad_epochs_gpu()` call which processes all epochs at once (not per-epoch), so the verbose warning should only print once if GPU fails, not repeatedly. We also check GPU availability upfront via `_should_use_gpu()` before even reaching this code, so this try/except is mainly a safety net for rare runtime issues that can't be detected statically.

autoreject/tests/test_backends.py
Ramdam17: Done!

autoreject/tests/test_backends.py
Ramdam17: I've completed the tests to check if:
- `get_reject_log()` works correctly with GPU backend
- `reject_log.labels` structure is correct
- `reject_log.bad_epochs` is populated properly  
- `reject_log.plot()` runs without errors
- Results are identical between numpy and torch backends

.flake8
Ramdam17: Done! Migrated all flake8 configuration from .flake8 to pyproject.toml.

autoreject/autoreject.py
Ramdam17: Reverted to the original single-line format (two-lines actually, because of the 88 chars rule)

autoreject/autoreject.py
Ramdam17: I tried to revert to original formatting everywhere, hope it will be more readable!

agramfort:
this PR is a mix of code reformatting and a new feature. I am all for reformatting using ruff but it should be done first so the PR with the actual new code is easier to review. Does that sound doable?
We should have pre-commit hooks setup here if it's not already the case.

autoreject/autoreject.py
jasmainak: @Ramdam17 I don't see the changes though. Maybe you have a git issue? Problem with a rebase or something?

PR #358, opened 2026-08-02
Feat/threshold search reproducibility

deep-introspection:


larsoner:
@deep-introspection I like the idea of adding a `torch` backend in principle! I would have thought maybe `array-api` would have been the way to go... or an issue to discuss or so before a big PR. But I might be missing some important context here!

deep-introspection:
@larsoner Sorry, I wanted to PR on @Ramdam17's fork! Everythign here is still in progress.

PR #359, opened 2026-09-01
Speed up AutoReject by hoisting loop invariants and batching epoch interpolation

larsoner:
I pressed Claude Fable 5 to find ways to speed up `autoreject` for MNE-BIDS-Pipeline, and these are some efficiency gains it found that don't change the results. I reviewed the changes and iterated a bit and everything looks reasonable to me!

- `_compute_thresh` moves computations that don't depend on the threshold
- Epoch repair via `_interpolate_bad_epochs` sped up by grouping epochs by bad-channel set and interpolated in one batched call (avoid a full Epochs deepcopy and interpolation-matrix computation *per epoch*, expensive)
- `_get_epochs_interpolation` computes peak-to-peaks once over the whole data array (avoiding an Epochs deepcopy per epoch)
- `_run_local_reject_cv` moves computations that are invariant out of the grid loop
- A few misc cleanups (e.g., one `np.ptp` in `_ChannelAutoReject.fit`, reusuing a `get_data()` in `_compute_thresholds`, single index in `_slicemean`)

I also checked memory consumption and it goes down slightly in this PR or is even with current `main` in these code paths.

I think we should merge and cut a quick release so people can benefit from the speed improvements!


larsoner:
... I'm going to bump to 3.11 min Python version as 3.10 is EOL in about a month (and way out of the SPEC0 support window)

larsoner:
6m41s to build the doc on the previous weekly run, down to 4m44s with these changes. I then parallelized that build the way I've done in MNE, and that took it to 2m27s :rocket:

jasmainak:
Beautiful @larsoner! Thanks so much. I am glad Fable didn't refuse to answer your questions ...

jasmainak:
There is an open PR: https://github.com/autoreject/autoreject/pull/347 to speed up autoreject even further ... you might have some thoughts on the approach and how it might impact MNE-BIDS-Pipeline.

PR #361, opened 2026-09-23
Fix switcher check

larsoner:
Should fix the warning banner, and modernize theme usage. I'll cut a 0.5.1 and that should be what fixes the deployed docs version numbering @cbrnr . This is just nice cleanups I figured were useful while looking into it


cbrnr:
Thanks @larsoner! So it was basically changing the `"version"` field from `"stable"` to `"0.5"`, right?

larsoner:
It was pushing a new release that used the `stable` deployment you added

cbrnr:
I'm seeing the unstable development banner for 0.5 (stable) (in addition to 0.6 (devel)).

Also, I would remove 0.2 and 0.3 unless you really need them (0.2 doesn't even have a version switcher).

larsoner:
I think that's just https://github.com/pydata/pydata-sphinx-theme/pull/2497 , so should be fixed by the time we cut another release (they're on 0.22.rc so should come out soon)

